PR #134, opened 2026-06-29
Fix istartswith / iendswith lookups being case-sensitive

rmortes:
## Summary

The `istartswith` and `iendswith` field lookups were **case-sensitive**, compiling to exactly the same SurrealQL as their case-sensitive counterparts `startswith` / `endswith`. Neither the field nor the bound value was lowercased, which contradicts the `i`-prefix convention used elsewhere in the ORM (`icontains`, `iendswith`, `ilike`, `iregex`).

For example, `name__istartswith="ali"` did **not** match `"Alice"`, because it compiled to: string::starts_with(name, $_f0) -- with $_f0 = "ali"

## Fix
In `surreal_orm/query_set.py`, `_render_condition` split `istartswith` and `iendswith` out of their shared branches and now lowercases **both** sides, mirroring the existing `icontains` implementation:

```
string::starts_with(string::lowercase(name), $_f0) -- with $_f0 = "ali"
string::ends_with(string::lowercase(name), $_f0) -- with $_f0 = "ce"
```

Both now also raise `TypeError` for non-string values, consistent with the other
case-insensitive string lookups.

## Tests
Followed a red-green approach: added tests that failed against the old behavior, then
applied the fix:
- **Unit** (`tests/test_unit.py`): assert the compiled SurrealQL lowercases both sides and
  is no longer byte-for-byte identical to the case-sensitive variant.
- **Integration** (`tests/test_query_operators_integration.py`): verify against a live
  SurrealDB (`name__istartswith="al"` matches `"Alice Smith"`, `name__iendswith="JONES"`
  matches `"Bob Jones"`), and that the case-sensitive variant with the same casing matches
  nothing.

All 141 unit tests and 24 operator integration tests pass.

### DIFF
--- a/src/surreal_orm/query_set.py
+++ b/src/surreal_orm/query_set.py
@@ -1006,13 +1006,29 @@ class QuerySet(Generic[T]):
             return f"{field_name} {op} {value}"
 
         # ── Function-based lookups (no SurrealQL operator equivalent) ─
-        # startswith / istartswith → string::starts_with()
-        if lookup_name in ("startswith", "istartswith"):
+        # startswith → string::starts_with(field, value)
+        if lookup_name == "startswith":
             return f"string::starts_with({field_name}, ${_bind(value)})"
 
-        # endswith / iendswith → string::ends_with()
-        if lookup_name in ("endswith", "iendswith"):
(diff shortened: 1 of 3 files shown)


EulogySnowfall (approved):
GJ! 👍🏼

PR #135, opened 2026-06-29
Fixed CREATE and UPDATE behaving differently with regard to permissions

rmortes:
## Summary

On a table protected by row-level `PERMISSIONS`, the write paths behaved inconsistently when SurrealDB denied a write:

- **CREATE** — `Model(**data).save()` on a new instance **raised** `SurrealDbError("Can't save data, no record returned.")`.
- **UPDATE / merge** — `.save()` on a persisted instance, or `.merge(...)`, on a row the user could not update was a **silent no-op**: it returned `self`, the row was unchanged, and no error was raised.

SurrealDB returns an *empty result* (zero affected rows) both when the target row is missing and when row-level permissions deny the write — it does not raise. The merge/update paths discarded that result, so a denied update was indistinguishable from a successful one. Application code relying on `merge()` / `save()` to enforce "the write happened" could silently lose data and report success.

## Fix

A denied/missing update now raises `SurrealDbError`, mirroring the create-path guard, across **every** non-deferred write path:

- plain `client.merge()` (persisted `save()` and `merge()`)
- the `SurrealFunc` raw-`UPDATE` path (`merge(field=SurrealFunc(...))`,
  `save(server_values=...)`)
- the complex-nested-data SET-clause path (`_execute_save_using_query`)
- WebSocket transactions (results are immediate)

The check is centralized in a single helper:

```python
def _raise_if_no_record_affected(self, result, thing, tx=None):
    if tx is not None and tx.defers_results:
        return
    if result is None or result.is_empty:
        raise SurrealDbError(_UPDATE_NO_RECORD_MSG.format(thing=thing))

```

## Transaction nuance (`defers_results`)

HTTPTransaction batches statements and returns an empty placeholder response from each operation until commit(). The real per-statement result only exists at commit time. A naïve is_empty check would therefore make every HTTP-transaction merge/save raise.

To distinguish a deferred placeholder from a genuinely empty (denied) result, a defers_results property was added to the transaction abstraction:

- BaseTransaction.defers_results → False (immediate results, default)
- HTTPTransaction.defers_results → True (buffered until commit)
- WebSocketTransaction inherits False
The guard skips deferred transactions, so HTTP-transaction behavior is unchanged.

## Known limitation

Denial inside an HTTP transaction remains undetectable at statement time: the real result is not known until commit(), and a denied UPDATE returns an empty result without error even there. Detecting it would require inspecting the commit() QueryResponse in the transaction layer, which is out of scope here. This is why the WS-transaction case is covered by unit tests rather than integration tests (the ORM uses an HTTP connection by default).

### Verification

Full suite: 2243 passed, 27 skipped, 0 failed
mypy (strict): 66 source files, 0 errors

### DIFF
--- a/src/surreal_orm/model_base.py
+++ b/src/surreal_orm/model_base.py
@@ -35,6 +35,15 @@ class SurrealDbError(Exception):
 
 logger = logging.getLogger(__name__)
 
+# Raised when an update/merge affects no records. SurrealDB returns an empty
+# result both when the target row does not exist and when row-level
+# PERMISSIONS deny the write; in either case the update did not happen, so we
+# surface it instead of silently no-opping (mirrors the create-path guard).
+_UPDATE_NO_RECORD_MSG = (
+    "Can't update data, no record affected for {thing} "
+    "(record may not exist or the write was denied by permissions)."
+)
+
 # Global registry of all SurrealDB models for migration introspection
 _MODEL_REGISTRY: list[type["BaseSurrealModel"]] = []
 
@@ -931,8 +940,10 @@ class BaseSurrealModel(BaseModel):
                 )
             variables.update(extra_vars)
 
-        if self._db_persisted and id is not None:
-            thing = format_thing(table, id)
+        is_update = self._db_persisted and id is not None
+        thing: str | None = None
+        if is_update:
+            thing = format_thing(table, id)  # type: ignore[arg-type]
             query = f"UPDATE {thing} SET {set_clause};"
         elif id is not None:
             thing = format_thing(table, id)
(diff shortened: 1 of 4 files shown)


src/surreal_orm/model_base.py
```diff
@@ -35,6 +35,15 @@ class SurrealDbError(Exception):
...
 logger = logging.getLogger(__name__)
 
+# Raised when an update/merge affects no records. SurrealDB returns an empty
+# result both when the target row does not exist and when row-level
+# PERMISSIONS deny the write; in either case the update did not happen, so we
+# surface it instead of silently no-opping (mirrors the create-path guard).
+_UPDATE_NO_RECORD_MSG = (
+    "Can't update data, no record affected for {thing} "
+    "(record may not exist or the write was denied by permissions)."
+)
```
EulogySnowfall, line 44: `ruff format` collapses this to a single line (it fits within the configured line length). This is the only thing failing CI:

```suggestion
_UPDATE_NO_RECORD_MSG = "Can't update data, no record affected for {thing} (record may not exist or the write was denied by permissions)."
```

EulogySnowfall (commented):
Thanks for this fix — the logic is solid and all unit tests pass. The **only** CI failure is a cosmetic `ruff format --check` violation in `src/surreal_orm/model_base.py` (the `_UPDATE_NO_RECORD_MSG` string is split across lines but fits within the line-length limit, so `ruff format` collapses it to a single line). No other job failed; `Integration Tests` was skipped only because it depends on `Lint & Type Check`.

Running `uv run ruff format` locally and pushing the result will make CI green. You can also just accept the committable suggestion below.

PR #166, opened 2026-08-23
fix(sdk): decode compact CBOR datetimes

### LINKED ISSUE #165: Datetimes decode as a tuple of ints outside typed model fields
`_cbor_tag_decoder()` only parses tag 12 when it carries an ISO 8601 string, but SurrealDB 3.x sends the compact `[seconds, nanoseconds]` form for every datetime. The decoder returns that pair unchanged.

Only model fields annotated `datetime` are not affected, because `_parse_datetime()` re-parses the pair afterwards.

Reproduction:
```python
class M(BaseSurrealModel):
    model_config = SurrealConfigDict(table_name="dm")
    id: str | None = None
    when: datetime | None = None
    payload: dict | None = None

ts = datetime(2025, 2, 13, 5, 0, tzinfo=UTC)
await M(id="a", when=ts, payload={"at": ts}).save()
rows = await M.raw_query("SELECT * FROM dm;")

type(rows[0]["when"])  # tuple, expected datetime
type(rows[0]["payload"]["at"])  # tuple, expected datetime
```

This affects `raw_query()`, datetimes nested inside object fields and any other result the ORM does not coerce per annotation.

System:
Ubuntu 24.04.4 LTS, Python 3.14.6, Surreal 3.3.0-beta.3+20260819.6dce5c8

aliciusschroeder:
SurrealDB 3.x sends tag 12 as a [seconds, nanoseconds] pair, not an ISO string. Only fields annotated datetime were rescued by the ORM; raw queries and datetimes nested in objects returned a tuple of ints.

Decoding the compact form makes `_parse_datetime()`'s array branch redundant for CBOR, though it is left in place for the JSON protocol.

Reproduction:

```pythonrepl
>>> import asyncio
>>> from datetime import UTC, datetime
...
... from surreal_orm import BaseSurrealModel, SurrealConfigDict, SurrealDBConnectionManager
...
... SurrealDBConnectionManager.set_connection("http://localhost:8000", "root", "root", "dev", "test")
...
... class M(BaseSurrealModel):
...     model_config = SurrealConfigDict(table_name="dm")
...     id: str | None = None
...     when: datetime | None = None
...     payload: dict | None = None
...
...
... ts = datetime(2025, 2, 13, 5, 0, tzinfo=UTC)
... await M(id="a", when=ts, payload={"at": ts}).save()
... rows = await M.raw_query("SELECT * FROM dm;")
```

Before:
```pythonrepl
>>> type(rows[0]["when"])
<class 'tuple'>
>>> type(rows[0]["payload"]["at"])
<class 'tuple'>
>>> rows
[{'id': RecordId(table='dm', id='a'), 'payload': {'at': (1739422800, 0)}, 'when': (1739422800, 0)}]
```

After:
```pythonrepl
>>> type(rows[0]["when"])
<class 'datetime.datetime'>
>>> type(rows[0]["payload"]["at"])
<class 'datetime.datetime'>
>>> rows
[{'id': RecordId(table='dm', id='a'), 'payload': {'at': datetime.datetime(2025, 2, 13, 5, 0, tzinfo=datetime.timezone.utc)}, 'when': datetime.datetime(2025, 2, 13, 5, 0, tzinfo=datetime.timezone.utc)}]
```

System:
Ubuntu 24.04.4 LTS, Python 3.14.6, Surreal 3.3.0-beta.3+20260819.6dce5c8

Closes #165


EulogySnowfall:
Thanks — good catch, and a well-scoped diagnosis. Confirmed and merging shortly.

**Verified.** I reproduced it against **3.2.4** (the stable release, not just the 3.3.0 beta you tested on), so this affects every SurrealDB version `main` supports:

```
before: {'payload': {'at': (1739422800, 0)}, 'when': (1739422800, 0)}
after:  {'payload': {'at': datetime(2025, 2, 13, 5, 0, tzinfo=utc)}, 'when': datetime(...)}
```

Your reading of the cause is right: tag 12 carries the compact `[seconds, nanoseconds]` pair, and `_parse_datetime()` was only rescuing values that happened to land in a field annotated `datetime`.

**One thing I fixed on top** (pushed to your branch as 72d8bea, hope that's alright):

`datetime.fromtimestamp()` raises on values SurrealDB accepts happily — its datetime range is far wider than Python's year 1–9999. Raising inside a cbor2 tag hook isn't local to the field: cbor2 re-raises it as `CBORDecodeError`, so a single unrepresentable datetime takes down the decode of the **whole** response. Before your change such a value came back as a harmless tuple, so it would have been a regression:

```sql
CREATE t SET d = <datetime>'+25252-01-01T00:00:00Z';
SELECT * FROM t;   -- cbor2.CBORDecodeError: error decoding semantic tag 12
```

Both branches now fall back to the raw tag value, which is exactly what `_parse_datetime()` already does for the same input. I also added the regression tests the change needed (`tests/sdk/test_cbor_protocol.py::TestCompactDatetimeDecoding`) — compact pair, sub-second precision, the cbor2 6.x tuple form, nested payloads, the ISO string path, and the out-of-range fallback.

**Verification:** `ruff`, `mypy` clean; unit suite green; `make test-integration` → 433 passed against SurrealDB 3.2.4.

Note the pin on `main` currently reads `3.3.0-beta.3` — that's a CI bug on my side (#163), not something you did. #167 restores it to 3.2.4 and stops the monitor pinning pre-releases.

Issue #147, opened 2026-07-27
SurrealDB 3.2.0 regression: sub-SELECT with ORDER BY + LIMIT breaks Subquery


EulogySnowfall:
Filed manually — the `SurrealDB Version Monitor` should have opened this automatically but its `create-failure-issue` job was unreachable (see #146).

## Summary

SurrealDB **3.2.0** regressed uncorrelated sub-SELECTs that combine `ORDER BY` **and** `LIMIT`. When such a subquery appears inside an outer `SELECT`'s projection or `WHERE` clause, it yields a result for only *some* outer rows and `NONE`/`[]` for the rest.

This breaks `Subquery` and is why the version monitor has failed daily since ~2026-07-15:

```
FAILED tests/test_subquery.py::TestSubqueryIntegration::test_filter_by_subquery_scalar - assert 0 >= 1
```

**This is an upstream SurrealDB bug, not an ORM bug.** `.surrealdb-version` is still pinned to `3.1.3`, which is unaffected — no user impact today. The blocker is that we cannot advance to 3.2.0.

## Reproduction

Standalone SurrealQL, no ORM involved. Seed:

```sql
CREATE orders:o1 SET total = 100.0, status = 'completed';
CREATE orders:o2 SET total =  50.0, status = 'pending';
CREATE orders:o3 SET total = 200.0, status = 'completed';
```

Then:

```sql
SELECT total, array::first(
  (SELECT VALUE total FROM orders WHERE status='completed' ORDER BY total DESC LIMIT 1)
) AS mx FROM orders;
```

| | result |
|---|---|
| **3.1.3** | `[{mx: 200, total: 100}, {mx: 200, total: 50}, {mx: 200, total: 200}]` ✅ |
| **3.2.0** | `[{mx: 200, total: 100}, {mx: NONE, total: 50}, {mx: NONE, total: 200}]` ❌ |

Consequently the filter the ORM generates returns nothing on 3.2.0:

```sql
SELECT * FROM orders WHERE total = array::first(
  (SELECT VALUE total FROM orders WHERE status='completed' ORDER BY total DESC LIMIT 1));
-- 3.1.3 → orders:o3      3.2.0 → []
```

## Isolating the trigger

Only `ORDER BY` **and** `LIMIT` *together* trigger it — either alone is fine:

| subquery clauses | 3.2.0 |
|---|---|
| neither | ✅ same value every row |
| `LIMIT 1` only | ✅ same value every row |
[...]

PR #174, opened 2026-08-23
fix(orm): coerce ForeignKey values to RecordId on write and in filters

### LINKED ISSUE #169: ForeignKey values are never coerced to RecordId on write or in filters
Split out of #162 (group 2), reported by @leosilberg. Tracking issue.

The read path converts — `_preprocess_db_record()` turns `RecordId` into a `"table:id"` string — but nothing converts the other way. So FK fields only actually work against plain `string` columns, which defeats the point of `record<>` / `REFERENCE`.

**Writes.** `save()` / `create` / `merge` send `model_dump()` output, so an FK holds a plain string that a `record<>` column rejects. The ORM knows which fields are FKs — the markers are on the annotation — so `_execute_save` should wrap `"table:id"` strings in `RecordId` for those fields.

**Filters — the worst of the three.** `.filter(fk="table:id")` binds a CBOR string, and SurrealDB never equates a string with a record. That returns **empty results rather than an error**: a silent wrong answer, which is strictly worse than a crash. The `QuerySet` has the model class and can coerce filter values on FK fields.

**`on_delete` vocabulary.** Confirmed by inspection:

```python
# fields/relation.py
on_delete: Literal["CASCADE", "SET_NULL", "PROTECT"] | None = None

# migrations/operations.py, both emitters
parts.append(f"ON DELETE {self.on_delete.upper()}")
```

SurrealDB's clause takes `CASCADE | UNSET | REJECT | IGNORE | THEN`, so `SET_NULL` and `PROTECT` pass straight through and generate invalid DDL. Preference is to **map** (`SET_NULL`→`UNSET`, `PROTECT`→`REJECT`) rather than change the literals — the Django vocabulary is the point of the API, and accepting both spellings is cheap.

Priority: the filter coercion first, as the only one of these that fails silently.

leosilberg:
Fixes #169

## 1. ForeignKey coercion — supported syntax

`ForeignKey` fields keep holding `"table:id"` strings in Python, but every write path (`save`, `merge`, `update`, `bulk_update`) and every whole-record filter lookup (`exact`, `in`, `not_in`, incl. `Q` objects) now converts the value to a `RecordId` at the wire boundary — so `record<>` columns accept the writes, and filters match instead of silently returning empty results.

A foreign key reference is now callable in **four interchangeable forms**, mixed freely in collections:

```python
# Assignment / save — all equivalent
Article(author=alice)                                  # model instance (Django-style)
Article(author="fk_authors:alice")                     # full "table:id" string
Article(author="alice")                                # bare ID → qualified via ForeignKey("Author")
Article(author=RecordId(table="fk_authors", id="alice"))

# Filters — object filtering, same four forms
Article.objects().filter(author=alice)
Article.objects().filter(author="alice")
Article.objects().filter(author__in=[alice, "bob", "fk_authors:carol", RecordId(...)])  # mixed
Article.objects().filter(author__not_in=["fk_authors:spam"])
Article.objects().filter(Q(author=alice) | Q(author="bob"))
Article.objects().filter(~Q(author="fk_authors:banned"))

# ManyToMany items accept the same forms
Member(groups=[group_obj, "groups:g2", RecordId(table="groups", id="g3")])

# bulk_update writes record links too
Article.objects().filter(title="Hello").bulk_update({"author": "bob"})
```

Deliberately untouched: string lookups (`startswith`, `contains`, `regex`, …), `isnull`, and explicit `$var` references bound via `.variables()`.

New public API: the reserved `instance.record_id` property (the Django `Model.pk` analog) — the record identity as a `RecordId`, ready to bind against `record<>` columns in `raw_query()` / `.variables()`, where explicit bindings deliberately pass through uncoerced; `str(instance.record_id)` gives the `"table:id"` string form. Plus `to_record_id()` / `record_link_to_str()` helpers in `model_base` next to the existing `_convert_record_id_to_string`. The JSON protocol encoder now renders `RecordId` as `"table:id"` so the fallback protocol keeps working.

## 2. `on_delete` DDL fix

Django's `on_delete` vocabulary passed through verbatim into migration DDL, generating invalid clauses — SurrealDB takes `CASCADE | UNSET | REJECT | IGNORE | THEN`. `AddField` / `AlterField` now map through the new `on_delete_to_surql()`:

| Django | SurrealDB |
|---|---|
| `SET_NULL` | `UNSET` |
| `PROTECT` | `REJECT` |
| `CASCADE` / native keywords | unchanged |

As preferred in the issue, the Django literals stay the API; both vocabularies are accepted.

## Tests

- **46 unit tests** (`tests/test_foreign_key_coercion_unit.py`) covering every form above on write, filter, and assignment paths, plus alias handling, unknown targets, and the JSON encoder.
- **11 integration tests** (`tests/test_foreign_key_coercion_integration.py`) against a real SCHEMAFULL `record<fk_authors>` column — rows are created server-side with raw SurrealQL so the filters are proven against real record values, not just ORM-written ones. Each fails without the coercion.
- `on_delete` mapping guarded in `tests/test_migrations_operations.py` and `tests/test_relations.py`.
- Full suite: 1893 unit tests pass, mypy strict clean (66 files), ruff clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/surreal_orm/fields/__init__.py
+++ b/src/surreal_orm/fields/__init__.py
@@ -27,6 +27,7 @@ from .relation import (
     is_graph_relation,
     is_many_to_many,
     is_relation_field,
+    on_delete_to_surql,
 )
 from .vector import VectorField, get_vector_info, is_vector_field
 
@@ -66,4 +67,5 @@ __all__ = [
     "is_graph_relation",
     "is_many_to_many",
     "is_relation_field",
+    "on_delete_to_surql",
 ]
--- a/src/surreal_orm/fields/relation.py
+++ b/src/surreal_orm/fields/relation.py
@@ -28,6 +28,7 @@ from pydantic import GetCoreSchemaHandler, GetJsonSchemaHandler
 from pydantic.json_schema import JsonSchemaValue
 from pydantic_core import CoreSchema, core_schema
 
+from surreal_orm.model_base import record_link_to_str
 from surreal_orm.utils import escape_record_id
 
 if TYPE_CHECKING:
@@ -130,14 +131,17 @@ class _ForeignKeyMarker:
                 break
 
         return core_schema.nullable_schema(
-            core_schema.str_schema(
-                metadata={
-                    "relation_type": "foreign_key",
-                    "to_model": to_model,
-                    "on_delete": on_delete,
-                    "related_name": related_name,
-                    "surreal_type": "record",
-                }
+            core_schema.no_info_before_validator_function(
+                record_link_to_str,
+                core_schema.str_schema(
+                    metadata={
+                        "relation_type": "foreign_key",
+                        "to_model": to_model,
+                        "on_delete": on_delete,
+                        "related_name": related_name,
+                        "surreal_type": "record",
+                    }
+                ),
             )
         )
 
@@ -192,9 +196,13 @@ class _ManyToManyMarker:
                 related_name = arg.related_name
                 break
 
-        # ManyToMany is represented as a list of record IDs (virtual field)
+        # ManyToMany is represented as a list of record IDs (virtual field).
+        # Each item accepts a model instance, RecordId, or "table:id" string.
         return core_schema.list_schema(
-            core_schema.str_schema(),
+            core_schema.no_info_before_validator_function(
+                record_link_to_str,
+                core_schema.str_schema(),
+            ),
             metadata={
                 "relation_type": "many_to_many",
                 "to_model": to_model,
@@ -371,6 +379,34 @@ def Relation(
         - followers: SELECT * FROM user:id<-follows<-User
     """
     return Annotated[list, _RelationMarker(edge, to, reverse)]
+
+
+def on_delete_to_surql(on_delete: str | None) -> str | None:
+    """
+    Translate an ``on_delete`` value into its SurrealDB ``ON DELETE`` keyword.
+
+    :func:`ForeignKey` takes Django's vocabulary (``CASCADE``, ``SET_NULL``,
+    ``PROTECT``) while SurrealDB accepts
+    ``CASCADE | UNSET | REJECT | IGNORE | THEN``.  SurrealDB's own keywords
+    pass through unchanged.
+
+    Args:
+        on_delete: The configured delete behavior, in either vocabulary
+
+    Returns:
+        The SurrealDB keyword, or ``None`` when *on_delete* is ``None``
+
+    Examples:
+        on_delete_to_surql("SET_NULL")  # "UNSET"
+        on_delete_to_surql("PROTECT")   # "REJECT"
+        on_delete_to_surql("CASCADE")   # "CASCADE"
+    """
+    if on_delete is None:
+        return None
+
+    # SurrealDB has no SET_NULL/PROTECT; the equivalents are UNSET and REJECT
+    normalized = on_delete.upper()
+    return {"SET_NULL": "UNSET", "PROTECT": "REJECT"}.get(normalized, normalized)
 
 
 def is_relation_field(field_type: Any) -> bool:
--- a/src/surreal_orm/migrations/operations.py
+++ b/src/surreal_orm/migrations/operations.py
@@ -10,6 +10,7 @@ from collections.abc import Callable, Coroutine
 from dataclasses import dataclass, field
 from typing import Any
 
+from ..fields.relation import on_delete_to_surql
 from ..types import FieldType
 
 
@@ -267,7 +268,7 @@ class AddField(Operation):
         if self.reference:
             parts.append("REFERENCE")
             if self.on_delete:
-                parts.append(f"ON DELETE {self.on_delete.upper()}")
+                parts.append(f"ON DELETE {on_delete_to_surql(self.on_delete)}")
 
         if self.comment:
             escaped_comment = self.comment.replace("'", "''")
@@ -398,7 +399,7 @@ class AlterField(Operation):
         if self.reference:
             parts.append("REFERENCE")
             if self.on_delete:
-                parts.append(f"ON DELETE {self.on_delete.upper()}")
+                parts.append(f"ON DELETE {on_delete_to_surql(self.on_delete)}")
 
         return " ".join(parts) + ";"
 
--- a/src/surreal_orm/model_base.py
+++ b/src/surreal_orm/model_base.py
@@ -13,6 +13,8 @@ from typing import (
 
 from pydantic import BaseModel, ConfigDict, PrivateAttr, model_validator
 
+from surreal_sdk.protocol.cbor import RecordId
+
 from . import signals as model_signals
 from .connection_manager import SurrealDBConnectionManager
 from .debug import _elapsed_ms, _log_query, _start_timer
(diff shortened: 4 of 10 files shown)


EulogySnowfall:
## Review

Thanks for this — the core of #169 is solidly addressed. `_to_record_link()` handles all four input forms, `$var` passthrough is preserved, string lookups and `isnull` are correctly left alone, `get_foreign_key_targets()` reads the marker off `FieldInfo.metadata` (right call — Pydantic strips `Annotated`), and the `on_delete_to_surql()` mapping is correct and tested. The circular-import dance between `fields.relation` and `model_base` works in both import paths.

I checked out the branch and verified locally: **1940 unit tests pass**, **mypy strict clean** (66 files), **ruff clean**.

Five findings below, each reproduced against a real model tree. Three are blocking.

### Two process notes first

- **CI has never run on this PR.** All three `ci.yml` runs on this branch are `action_required` (fork-contributor approval pending), so nothing here was machine-verified before I ran it locally. The workflow run needs approving so CI actually reports on this PR.
- **The branch is 6 commits behind main** and touches `operations.py` near #168's `AlterField` OVERWRITE fix. The merge-base diff is clean — no actual conflict — but please **rebase on main and re-verify**, so #168 can't be silently undone.

---

### 1. Regression: `filter(fk__in={...})` with a `set` now raises

`_coerce_filter_value` rebuilds the collection with `type(value)(...)`, but `RecordId` is an unhashable dataclass:

```
TypeError: cannot use 'RecordId' as a set element (unhashable type: 'RecordId')
```

This worked before the PR. The tests cover `list` and `tuple` but not `set`.

Since `IN` takes an array regardless, always build a `list`:

```python
        table = targets[field_name]
        if isinstance(value, (list, tuple, set)):
            # Always a list: RecordId is an unhashable dataclass, so rebuilding
            # a set would raise, and IN takes an array either way.
            return [_to_record_link(item, table) for item in value]
        return _to_record_link(value, table)
```

### 2. Aliased ForeignKeys are still broken in filters

This is the failure mode #169 singles out as "the worst of the three" — the one that returns a wrong answer instead of an error.

The write path resolves aliases (`_coerce_foreign_keys` handles them); the read path keys off the Python field name only, while filters address **database columns**:

| filter spelling | generated SQL | bound variable | |
|---|---|---|---|
| no alias | `author = $_f0` | `RecordId(...)` | ✅ |
| by alias (`author_id=`) | `author_id = $_f0` | `'fk_authors:alice'` | ❌ plain string |

So an aliased FK is written as a record link but filtered as a string → empty results, no error. `bulk_update` has the same defect (it also keys on `get_foreign_key_targets()`).

Suggested fix — a column-keyed accessor in `model_base.py`, used by both query paths:

```python
    @classmethod
    def get_foreign_key_columns(cls) -> dict[str, str | None]:
        """
        Map each foreign-key *column* name to its target table.

        Filters and bulk writes address database columns, so an aliased
        ``ForeignKey`` must be resolvable under its alias too.
        """
        columns: dict[str, str | None] = {}
        for field_name, table in cls.get_foreign_key_targets().items():
            columns[field_name] = table
            alias = cls.model_fields[field_name].alias
            if alias:
                columns[alias] = table
        return columns
```

then swap `get_foreign_key_targets()` → `get_foreign_key_columns()` in `_coerce_filter_value` and in `bulk_update`.

I applied exactly this plus fix 1 locally: aliased filters bind a `RecordId`, the set case no longer raises, **1940 tests still pass and mypy stays clean**. Happy to push it if you'd rather I did.

> Out of scope, but noting it: filtering by the *Python* field name on an aliased model emits `author = ...` rather than the real column `author_id`. That's pre-existing — `_render_condition` never translated aliases — not something this PR introduced. Worth a separate issue.

### 3. `upsert()` and `bulk_upsert()` were missed

Both still bind FK values as raw strings, so the bug this PR fixes for `save`/`merge`/`update`/`bulk_update` persists there and a `record<>` column will reject the statement.

- `upsert()`: `defaults` and `on_conflict` are caller-supplied and keyed by column name → route values through `_to_record_link()` using `get_foreign_key_columns()`.
- `bulk_upsert()`: simplest fix is to reuse the existing method right after the dump, since it already handles aliases:

```python
            data = instance.model_dump(exclude_unset=True, by_alias=True)
            data.pop("id", None)
            data = instance._coerce_foreign_keys(data)
```

(I haven't implemented these two — proposal only.)

### 4. `update()` and `merge()` now disagree on signal payloads

`update()` coerces first and passes the coerced dict as `update_fields`, so handlers receive `RecordId` objects; `merge()` deliberately keeps the uncoerced `data_set`. `merge()`'s choice is the right one — signals are a public API and shouldn't leak wire types. Suggest `update()` keep an uncoerced dict for the three signal sends and pass the coerced one only to `tx.merge()` / `client.merge()`.

### 5. Assigning an unsaved instance gives a confusing error

```python
Article(author=Author(name="Alice"))
# Input should be a valid string [type=string_type, input_type=Author]
```

Rejecting an unsaved reference is defensible (Django does too), but the message doesn't explain the cause. A saved instance works fine. Either raise something explicit from the validator ("cannot reference an unsaved <Model>; save it first"), or note the constraint in the PR description, which currently advertises instance assignment without the caveat.

---

Findings 1–3 should land before merge; 4 is consistency, 5 is polish.

leosilberg:
## Review findings — resolved

Thanks for the thorough pass, and for taking the time to reproduce each one against a real model tree. All five are addressed in f80920c, on top of a rebase. Point-by-point below, including two places I deviated from your suggestion.

### Process notes

**Rebase** — done. The branch is now on `39e01e8` (0 behind / 2 ahead before this commit). The rebase was conflict-free, and I checked #168 specifically: all three `DEFINE FIELD OVERWRITE` sites are intact, and this branch's only remaining touch on `operations.py` is the two `on_delete_to_surql()` call sites plus the import — different lines, no overlap. Nothing of #168 was undone.

**CI** — still `action_required` on the new head, so the workflow run needs approving whenever you get a chance. Everything below is local verification only.

---

### 1. `filter(fk__in={...})` with a `set` — fixed

Rebuilt as a `list` exactly as you suggested. Confirming your diagnosis was precisely scoped: `set` was the *only* broken collection type. A `tuple` already came out as a `list`, because it gets normalized downstream when the `IN` value is bound — so the fix is genuinely about `RecordId` being unhashable, nothing more.

### 2. Aliased ForeignKeys in filters — fixed

Took your `get_foreign_key_columns()` verbatim and swapped it in at both `_coerce_filter_value` and `bulk_update`. Aliased filters now bind a `RecordId`, in the plain, `__in`, and `Q` spellings alike.

Your out-of-scope note still stands and is untouched here: filtering by the *Python* field name on an aliased model still emits `author = ...` rather than the real column `author_id`. Confirmed still pre-existing on this branch. Happy to open a separate issue for it.

### 3. `upsert()` and `bulk_upsert()` — fixed

This was proposal-only on your side, so it is the one part with no prior implementation behind it and probably deserves the closest look.

- `upsert()` — `defaults` and `on_conflict` route through `_to_record_link()` via `get_foreign_key_columns()`, at all three binding sites. `SurrealFunc` values still inline uncoerced.
- `bulk_upsert()` — reuses `_coerce_foreign_keys()` right after the dump, as you suggested.

**Deviation:** I also coerced `bulk_upsert()`'s `on_conflict`, which your note did not cover. It has the identical defect to `upsert()`'s — caller-supplied and keyed by column — so leaving it would have fixed the batch's row data while its conflict clause still bound strings. Say the word if you would rather that were a separate change.

### 4. `update()` / `merge()` signal payloads — fixed

`update()` now keeps `data` uncoerced for the three signal sends and passes a separate `wire_data` to `tx.merge()` / `client.merge()`, mirroring `merge()`'s existing split. `_log_query` logs `wire_data`, since that is what actually went over the wire. Agreed that `merge()`'s choice was the right one — signals should not see wire types.

### 5. Unsaved instance — fixed, but placed differently

**Deviation, and the one I would most like your opinion on.** You suggested raising from the validator. I put the guard in `record_link_to_str()` instead, because the field validators are not the only caller — `_to_record_link()` calls it too, so the write and filter paths were *silently binding an unsaved instance into the query as a raw model object*. One guard in the shared helper covers all three:

| path | before | after |
|---|---|---|
| `Article(author=unsaved)` | `Input should be a valid string` | `cannot reference an unsaved Author; save it first` |
| `Member(tags=[unsaved])` | same unhelpful message | same clear message |
| `filter(author=unsaved)` | bound a model object into the query | raises |

The tradeoff is real and yours to accept or reject: `record_link_to_str()` is documented as total ("values that are neither are returned unchanged") and the PR advertises it as a public helper, so it can now raise. I added a `Raises:` section to its docstring. Inside a validator the `ValueError` surfaces as a normal `ValidationError`; on the write path it propagates as `ValueError`. The check is ordered after both `RecordId` returns and gated on `callable(getattr(value, "get_id", None))`, so the existing passthrough for unrelated objects carrying a plain `record_id` attribute is preserved — there is a test pinning that. If you would rather keep the helper total, moving it back into a validator wrapper is a small change.

---

### Verification

- **1943 unit tests pass** (was 1923), **mypy strict clean** (66 files), **ruff check + format clean**.
- `tests/test_foreign_key_coercion_unit.py` grew 49 → 69. I checked the new tests against reverted `src/`: **14 of the 20 fail without the fix**; the other 6 are deliberate guards (string lookups still untouched, `update()` still writes a `RecordId`, saved instance and `None` still accepted, unrelated-object passthrough).
- **Not covered:** the new tests are unit-level with mocked clients. I did not add integration tests for the `upsert()` / `bulk_upsert()` paths, so those two are proven against generated SQL and bound variables rather than a live `record<>` column. Worth flagging given finding 3 had no prior implementation.
- The 29 failures in `tests/test_surrealdb_monitor.py` on my machine are pre-existing and Windows-only — those tests shell out to `bash -c` and hit Windows' `sort.exe`, which has no `-V`. Identical failures on pristine `main`, so unrelated to this branch, but it does mean that file is the one thing I could not self-verify locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

EulogySnowfall (approved):
## Re-review — all five findings verified fixed

Checked out `f80920c` and verified each finding against a live SurrealDB 3.2.4, rather than taking the summary at its word.

**Rebase confirmed.** `0` behind / `3` ahead of `main` (`39e01e8`). I re-checked #168 specifically: the diff touches `operations.py` only at the two `on_delete_to_surql()` call sites plus the import — all three `DEFINE FIELD OVERWRITE` sites are intact. Nothing of #168 was undone.

### Findings

| # | Fix | Verified |
|---|---|---|
| 1 | `set` rebuilt as `list` in `_coerce_filter_value` | ✅ live |
| 2 | `get_foreign_key_columns()` at both `_coerce_filter_value` and `bulk_update` | ✅ live |
| 3 | `upsert()` binds via `_to_record_link`; `bulk_upsert()` reuses `_coerce_foreign_keys` | ✅ live |
| 4 | `update()` keeps `data` uncoerced for signals, passes `wire_data` to the wire | ✅ code + test at `test_foreign_key_coercion_unit.py:599` |
| 5 | Guard in `record_link_to_str()` | ✅ runtime |

**On finding 3 — the gap you flagged is now closed.** You were right that this was the part deserving the closest look: it had no prior implementation, and your own note said it was proven against generated SQL and bound variables rather than a live `record<>` column. I wrote seven integration probes against a real SCHEMAFULL `record<pk_authors>` column covering `upsert()` defaults, `upsert()` `on_conflict`, `bulk_upsert()` row data, `bulk_upsert()` `on_conflict`, aliased-FK upsert, the `set` filter, and the aliased-FK filter.

**All seven pass on `f80920c`; all seven fail with `src/` reverted to `main`.** So the coercion is genuinely load-bearing on those paths, server-side, not just in the generated SQL.

### Deviations — both accepted

- **`bulk_upsert()`'s `on_conflict`.** Correct call, and I'd have asked for it if you hadn't done it. Fixing the batch's row data while its conflict clause still bound strings would have left exactly half the defect in place. My probe covers it.
- **Guard in `record_link_to_str()` rather than the validator.** Your reasoning is better than my suggestion. The validator is not the only caller, and the write and filter paths were silently binding an unsaved model object into the query — a validator-only guard would have left that. I accept the loss of totality: the `Raises:` docstring is there, the ordering after both `RecordId` returns is right, and the `callable(getattr(value, "get_id", None))` gate preserves the unrelated-object passthrough with a test pinning it. Message reads well: `cannot reference an unsaved Author; save it first`.

### Verification on this machine

- **1998 unit tests pass**, 0 failures.
- **430 integration tests pass** against SurrealDB 3.2.4 (1 skip, 1 known upstream xfail for #147).
- **mypy strict clean** — 66 files, run exactly as CI does (`uv sync --group lint`).
- **ruff check + format clean** — 149 files.

Your 29 `test_surrealdb_monitor.py` failures were indeed Windows-only: they pass here on Linux. That file is now covered.

### Process

CI had never run on this PR — every run was `action_required` pending fork approval, including on `f80920c`. I've approved the run on the current head, so CI reports on this PR for the first time. The diff touches `src/` and `tests/` only, no workflow or build-config changes.

Two follow-ups, neither blocking and both already agreed as out of scope: the pre-existing alias-rendering bug (filtering by the Python field name on an aliased model emits `author = ...` rather than the real column), and folding the `upsert()`/`bulk_upsert()` integration probes into the suite so that path stays covered.

Approving.

PR #179, opened 2026-09-01
fix(migrations): introspect relation markers, virtual fields and nullability

### LINKED ISSUE #170: Migration introspection drops relation markers, virtual fields and nullability
Split out of #162 (group 3), reported by @leosilberg. Tracking issue.

Four related defects in the model → `SchemaState` → operation path:

**`_introspect_field` never looks at relation markers.** It does not unwrap `Annotated`, so every `ForeignKey` and `ReferencesField` falls through `_map_type` to `any`. It should emit `record<target>` (nullable, `reference=True`, mapped `on_delete`). Everything downstream — `FieldState`, the operations, the DB-side `define_parser` — already carries those attributes; this one producer is the only thing that never fills them.

**`ManyToMany` / `Relation` introspect as columns typed `any`.** They are virtual — their own core-schema metadata says `surreal_type: "virtual"` — and should be excluded from DDL entirely.

**Nullability is dropped at the diff boundary.** `FieldState.nullable` exists and participates in equality, but `AddField` / `AlterField` have no `nullable` parameter and their `forwards()` never wraps `option<>`. Only `define_table()` does that wrapping, so generated migrations silently lose optionality.

**Inconsistent diff branches.** The "add field to an existing table" branch in `state.py` builds `AddField` without `reference` / `on_delete`, while the new-table branch includes them.

Depends on #169 for the `on_delete` mapping.

EulogySnowfall:
Fixes #170

Four related defects in the model → `SchemaState` → operation path.

### 1. `_introspect_field` never looked at relation markers

It did not unwrap `Annotated`, so every `ForeignKey` and `ReferencesField` fell through `_map_type` to `any` — this producer was the only thing that never filled the `reference` / `on_delete` attributes that `FieldState`, the operations and the DB-side `define_parser` all already carried.

Both now introspect as record links:

| Declaration | Generated DDL |
| --- | --- |
| `author: ForeignKey("User")` | `TYPE option<record<users>> REFERENCE ON DELETE CASCADE` |
| `books: ReferencesField["books", "CASCADE"]` | `TYPE option<array<record<books>>> REFERENCE ON DELETE CASCADE` |

The target table is resolved through `_resolve_target_table()` (already written for #169), so `ForeignKey("User")` finds the model's configured `table_name`. An unresolvable target stays an untyped `record` rather than an invented table name. The marker is read from the `Annotated` hint and from `FieldInfo.metadata`, since `_introspect_model` falls back to `FieldInfo.annotation` when `get_type_hints` fails and Pydantic strips the metadata off that attribute.

### 2. `ManyToMany` / `Relation` introspected as columns typed `any`

They are virtual — their own core schema says `surreal_type: "virtual"` — and are now excluded from DDL entirely. Filtering them in `_introspect_model` removes them from `CreateTable`, `AddField`, the diff and `define_table()` at once.

### 3. Nullability was dropped at the diff boundary

`AddField` / `AlterField` had no `nullable` parameter and their `forwards()` never wrapped `option<>`, so only `define_table()` kept optionality and generated migrations silently lost it. Both take `nullable` now, defaulting to `False` so hand-written migrations are unaffected; `AlterField` also takes `previous_nullable` so a rollback restores the optionality the column actually had. `define_table()` passes the flag instead of pre-wrapping the type itself, removing the duplicated wrapping logic. The wrap is skipped when the type already carries its own optionality (`option<T>`, or a union such as `none | T`).

### 4. Inconsistent diff branches

The "add field to an existing table" branch built `AddField` without `reference` / `on_delete` while the new-table branch included them. All three call sites now forward the full `FieldState`.

---

### ⚠️ Behavior change

`on_delete` used to be decorative: the column landed as `any` with no `REFERENCE` clause, so nothing was enforced. It is now enforced by SurrealDB, which means a `ForeignKey` left at the default `on_delete="CASCADE"` makes deleting the target delete the referencing record.

Because the `Literal` only accepted `CASCADE | SET_NULL | PROTECT`, there was no way to express "do nothing". It is widened to SurrealDB's vocabulary as well (`UNSET`, `REJECT`, `IGNORE`, `THEN`) and exported as the public `OnDelete` alias — `on_delete="IGNORE"` is the escape hatch for a link that should stay decorative. `on_delete_to_surql()` already translated both vocabularies since 0.32.7; only the typing blocked it.

This does not touch graph dynamics. `REFERENCE ON DELETE` is a `DEFINE FIELD` clause on the `record<>` column that declares it; `RELATE` edges live in their own tables with their own `in` / `out` fields, and defect 2 means `ManyToMany` / `Relation` now generate no DDL at all.

### ⚠️ Expect one-time `REMOVE FIELD` statements on the first diff after upgrading

Before this PR, `ManyToMany` and `Relation` introspected as real `any` columns, so existing databases physically carry those `DEFINE FIELD` entries. The model side now omits them as virtual, so the **first** `schema_diff` / `makemigrations` after upgrading emits `REMOVE FIELD groups ON users;` and friends, alongside the `any -> option<record<...>>` AlterFields.

No data is lost — those columns never held anything the ORM read, since graph edges live in their own tables — but the statements are auto-generated and should be reviewed rather than applied blind. Same for the `ON DELETE` clause now appearing on foreign keys (see the behavior change above).

**This needs a CHANGELOG / migration-guide note in the release PR**, per the release-checklist convention that docs land there rather than here.

### Side effect

A foreign key introspected as `any` while the database reported `record<...>`, so `schema_diff()` proposed a phantom `AlterField` on every run. An integration test now asserts the round-trip is stable; it was confirmed to fail (producing exactly that phantom `AlterField`) when the producer and the parser are made to disagree.

### Verification

| Command | Result |
| --- | --- |
| `uv run ruff check src/ tests/` | `All checks passed!` |
| `uv run python -m mypy src/` (via `uv sync --group lint`) | `Success: no issues found in 66 source files` |
| `make test` (unit) | exit 0 — 2004 passed, 26 skipped, 0 failures |
| `make test-integration` (SurrealDB 3.2.4) | exit 0 — 454 passed, 1 skipped, 1 xfail (upstream #147), 0 failures |

New tests: `tests/test_issue_170_unit.py` (31 cases across the four defects and the ten review findings) and `tests/test_issue_170_integration.py` (8 cases, including SurrealDB actually enforcing `ON DELETE CASCADE`, `INFO FOR TABLE` reporting no column for virtual fields, and a `SET_NULL` foreign key whose round-trip converges).

### Review round

The automated review found ten defects; every one reproduced by execution before being fixed, and all ten are addressed in `9d83c28`. Summary in the thread replies; the two that changed existing tests:

- `tests/test_phase2_beta.py::test_generates_references_field_with_on_delete` asserted the defective `inspectdb` mapping (a scalar `record<users>` generating `ReferencesField`). Corrected to expect `ForeignKey`, and the `array<record<T>>` + `ON DELETE` case it left uncovered was added.
- `test_foreign_key_carries_reference_and_on_delete` asserted `on_delete == "SET_NULL"`; `FieldState` now stores SurrealDB's `UNSET`, which is what makes the diff converge.

### Not included

Docs (`CHANGELOG.md`, `README.md`, `CLAUDE.md`) follow the release-checklist convention and belong in the release PR, not here. The behavior change above needs a `CHANGELOG` entry when that lands.


EulogySnowfall (commented):
Automated review (Claude Code, high effort — 8 finder angles + adversarial verification pass; every finding below was reproduced by executing the code, not just read). 8 inline comments follow; two findings anchor outside the diff so they live here.

**[correctness] `inspectdb` round-trip turns a scalar FK into `ReferencesField` → destructive scalar→array `AlterField`** — `src/surreal_orm/migrations/model_generator.py:200` (untouched by this PR, but newly reachable): `ModelCodeGenerator` maps *every* `reference=True` field to `ReferencesField["table"]`, including the scalar `record<T> REFERENCE` columns this PR's ForeignKey DDL now produces. Reproduced end-to-end: `TYPE option<record<users>> REFERENCE ON DELETE CASCADE` → generator emits `owner: ReferencesField["users", "CASCADE"]` → re-introspects as `array<record<users>>` → `SchemaState.diff` proposes `AlterField record<users> -> array<record<users>>` on a schema that was never edited. Scalar `record<T>` with `reference=True` should generate `ForeignKey(...)`; reserve `ReferencesField` for `array<record<T>>`.

**[migration-safety] First post-upgrade diff auto-proposes `REMOVE FIELD` for legacy M2M/Relation columns** — `src/surreal_orm/migrations/state.py:527`: pre-PR, `ManyToMany`/`Relation` fields introspected as real `any` columns, so existing databases physically carry those `DEFINE FIELD` entries; now the model side omits them as virtual, so the first `schema_diff`/`makemigrations` after upgrading emits `REMOVE FIELD groups ON users;` etc. (verified by running the introspector at the parent commit and diffing against head). No data is lost — the columns were virtual — but unreviewed auto-generated `REMOVE FIELD` alongside the `any -> option<record<...>>` AlterFields will surprise users. A CHANGELOG/migration-guide note telling users to expect and review these one-time statements would cover it.

src/surreal_orm/migrations/introspector.py
```diff
@@ -200,6 +256,47 @@ def _introspect_field(
...
+        else:
+            target = marker.table
+            field_type = FieldType.ARRAY.generic(FieldType.RECORD.generic(target))
+
+        return FieldState(
+            name=name,
+            field_type=field_type,
+            nullable=True,
+            reference=True,
+            on_delete=marker.on_delete,
```
EulogySnowfall: **[correctness — confirmed by execution] `on_delete` vocabulary mismatch → perpetual phantom `AlterField` for every non-CASCADE strategy.**

The model-side `FieldState` stores the Django vocabulary verbatim (`"SET_NULL"`/`"PROTECT"`), but `parse_define_field` returns what SurrealDB reports back (`"UNSET"`/`"REJECT"`), and `FieldState.__eq__` (state.py:58) compares the raw strings. Reproduced: `ForeignKey("User", on_delete="SET_NULL")` introspects as `on_delete='SET_NULL'`, the DDL emits `ON DELETE UNSET`, parsing that DDL yields `on_delete='UNSET'` — the states compare unequal, so every `schema_diff`/`makemigrations` proposes the same `AlterField` forever and applying it never converges. This is the exact phantom-diff symptom the PR fixes, surviving for every value except `CASCADE` — the one spelling identical in both vocabularies, and the only one the new integration test exercises.

Fix: normalize here via `on_delete_to_surql(marker.on_delete)` (or normalize in `FieldState.__eq__`), and add a `SET_NULL` case to `test_schema_diff_is_empty_after_define_table`.

src/surreal_orm/migrations/operations.py
```diff
@@ -407,7 +434,7 @@ def backwards(self) -> str:
         if not self.previous_type:
             return ""
 
-        normalized_prev_type = _normalize_field_type(self.previous_type)
+        normalized_prev_type = _apply_nullable(_normalize_field_type(self.previous_type), self.previous_nullable)
```
EulogySnowfall, line 518: **[correctness — confirmed by execution] `backwards()` silently strips the `REFERENCE` clause on rollback.**

`AlterField` has no `previous_reference`/`previous_on_delete` fields and `backwards()` never emits a `REFERENCE` clause. Reproduced: `AlterField(reference=True, on_delete="CASCADE", previous_type="record<users>", previous_nullable=True)` — `forwards()` emits `...TYPE record<users> REFERENCE ON DELETE CASCADE;` but `backwards()` emits `...TYPE option<record<users>>;` with no REFERENCE at all. Pre-existing in `backwards()`, but this PR is what first makes `reference=True` FieldStates (and thus these AlterFields) reachable from `makemigrations` — so rolling back any FK-field alter now silently disables referential integrity with no error.

Related, same block: state.py's alter branch also never forwards `previous_flexible`/`previous_readonly` (which `backwards()` honors), so rollback of a FLEXIBLE/READONLY field drops those clauses too.

src/surreal_orm/fields/relation.py
```diff
@@ -35,6 +35,25 @@ class User(BaseSurrealModel):
...
+#: ``ON DELETE`` clause, so deleting the target leaves this record alone.
+#: :func:`on_delete_to_surql` maps between the two.
+OnDelete = Literal[
+    "CASCADE",
+    "SET_NULL",
+    "PROTECT",
+    "UNSET",
+    "REJECT",
+    "IGNORE",
+    "THEN",
```
EulogySnowfall: **[correctness — confirmed by execution] `"THEN"` is accepted but there is no channel for its required expression — the emitted DDL is invalid.**

`on_delete_to_surql("THEN")` passes it through, and `AddField(field_type="record<users>", reference=True, on_delete="THEN").forwards()` emits `... REFERENCE ON DELETE THEN;` — SurrealDB's grammar requires `ON DELETE THEN <expression>`, so this is a parse error. Nothing in `ForeignKey`, `_ForeignKeyMarker`, `AddField`/`AlterField` or the introspector carries the expression, yet the docstring above claims "the custom clause configured on the field". `ForeignKey("User", on_delete="THEN")` type-checks and introspects cleanly, then `migrate()` fails with `MigrationStatementError` mid-migration — leaving partial state, since migrations aren't transactional.

Either drop `"THEN"` from the Literal or add an expression parameter alongside it.

src/surreal_orm/migrations/operations.py
```diff
@@ -59,6 +59,30 @@ def _normalize_field_type(field_type: FieldType | str) -> str:
...
+    Args:
+        field_type: The normalized SurrealDB type
+        nullable: Whether the column accepts NONE
+
+    Returns:
+        The type, wrapped in ``option<>`` when that is both needed and absent
+    """
+    if not nullable:
+        return field_type
+    if field_type.lower().startswith("option<") or "|" in field_type:
```
EulogySnowfall: **[correctness — confirmed by execution] `"|" in field_type` treats *any* union as already optional.**

Only unions containing `none` are optional. Reproduced: `AddField(field_type="int | string", nullable=True).forwards()` emits `TYPE int | string` — a non-nullable column despite the explicit flag, silently. Worse, the loop: `parse_define_field("TYPE option<string | int>")` yields `FieldState(field_type="string | int", nullable=True)`; re-emitting drops the wrap; re-parsing reads `nullable=False`; `diff()` proposes the same `AlterField` on every run, and each apply re-drops nullability — never converges.

Guard on a `none` member (matching `define_parser.py:182`'s detection: `option<`, `| null`, `none |`) instead of any pipe character — or normalize `option<>` into the `nullable` flag once at the operation boundary.

src/surreal_orm/migrations/introspector.py
```diff
@@ -200,6 +256,47 @@ def _introspect_field(
...
+        # so importing it at module scope would be circular.
+        from ..model_base import _resolve_target_table
+
+        if isinstance(marker, _ForeignKeyMarker):
+            target = _resolve_target_table(marker.to)
+            # An unresolved target stays an untyped `record` rather than an
+            # invented table name — a wrong table would be rejected on write.
+            field_type = FieldType.RECORD.generic(target) if target else FieldType.RECORD.value
+        else:
+            target = marker.table
```
EulogySnowfall: **[correctness — confirmed by execution] `ReferencesField` target used verbatim (and lowercased) while the FK branch resolves — wrong-table DDL.**

`_ReferencesMarker.__init__` (references.py:46) lowercases `table`, and this branch skips `_resolve_target_table`. Reproduced: `ReferencesField["I170Author"]` (whose model maps to table `CamelAuthors`) introspects to `array<record<i170author>>` — a nonexistent table baked into DDL — while `ForeignKey("I170Author")` on the same model correctly yields `record<CamelAuthors>`. Writes to the column are then rejected by the `record<>` type with no hint at the definition-time mistake. The lowercasing is pre-existing, but this PR is what first emits the target into `DEFINE FIELD`.

Route this branch through `_resolve_target_table` with the same untyped-`record` fallback the FK branch uses.

src/surreal_orm/migrations/introspector.py
```diff
@@ -6,20 +6,62 @@
...
+    the ``Annotated`` metadata off that attribute.
+
+    Args:
+        type_hint: The field's type annotation
+        field_info: Pydantic FieldInfo for the field
+
+    Returns:
+        The marker instance, or None when the field carries none
+    """
+    if get_origin(type_hint) is Annotated:
```
EulogySnowfall: **[correctness — confirmed by execution, reduced severity] A union-wrapped annotation misses the marker: `author: ForeignKey("User") | None = None` regresses to `any`.**

Wrapping in `| None` makes `get_origin()` a Union (not `Annotated`), and Pydantic does not lift metadata out of a union member into `FieldInfo.metadata` — so both lookups miss and the field introspects as `field_type=any, reference=False, nullable=False`: the #170 symptom resurfacing, while runtime validation still passes. Mitigating: the documented spelling `author: ForeignKey("User") = None` works correctly (the marker's inner type already includes `None`), so only the redundant outer `| None` — which mypy silently accepts since `ForeignKey(...)` is `Any` — triggers it.

Unwrapping a top-level Union before the Annotated check (or a docs note) closes it. Side note: the Annotated-args loop here is otherwise redundant — Pydantic always materializes top-level Annotated metadata into `FieldInfo.metadata`, so the second loop alone suffices.

src/surreal_orm/migrations/state.py
```diff
@@ -512,9 +516,11 @@ def diff(self, target: "SchemaState") -> list["Operation"]:
                                 value=field_state.value,
                                 reference=field_state.reference,
                                 on_delete=field_state.on_delete,
+                                nullable=field_state.nullable,
                                 previous_type=current_field.field_type,
                                 previous_default=current_field.default,
                                 previous_assertion=current_field.assertion,
+                                previous_nullable=current_field.nullable,
```
EulogySnowfall: **[simplification] The `FieldState` → operation kwarg expansion is now hand-copied in four places — the exact copy-drift shape defect 4 of #170 had.**

Both diff branches here, `model_base.define_table`, and `FieldState.__eq__` each enumerate the attribute list by hand; this PR fixes one drift (`nullable`) by adding a fifth manual thread-through, and a live instance remains on this very line's block: `previous_flexible`/`previous_readonly` exist on `AlterField` and are honored by `backwards()`, but are never forwarded here. The next `FieldState` attribute must be remembered in 4+ sites or #170 recurs a third time.

A single `AddField.from_field_state(table, fs)` classmethod (plus a `previous_*` builder for `AlterField`) wires each attribute once.

src/surreal_orm/fields/relation.py
```diff
@@ -35,6 +35,25 @@ class User(BaseSurrealModel):
     pass
 
 
+#: Accepted ``on_delete`` strategies.
+#:
+#: Django's vocabulary (``CASCADE``, ``SET_NULL``, ``PROTECT``) is the ORM's
+#: public API; SurrealDB's own keywords (``UNSET``, ``REJECT``, ``IGNORE``,
+#: ``THEN``) are accepted too and pass through untranslated.  ``IGNORE`` is the
+#: escape hatch for a purely decorative link: it emits ``REFERENCE`` with no
```
EulogySnowfall: **[docs — confirmed by execution] The `IGNORE` description doesn't match the code.**

This says IGNORE "emits `REFERENCE` with no `ON DELETE` clause", but the truthy `if self.on_delete:` gate in operations.py (AddField:296, AlterField:429) emits `REFERENCE ON DELETE IGNORE`. That DDL is valid and semantically equivalent, so the code is fine — the docstring's mechanism claim is what's wrong, and anyone extending the parser/diff who trusts it (e.g. treating an absent clause as meaning IGNORE) will mishandle round-trips. Note that actually suppressing the clause instead would trip the raw `on_delete` comparison in `FieldState.__eq__` (see the vocabulary-mismatch comment).

src/surreal_orm/migrations/introspector.py
EulogySnowfall: Confirmed by execution: `ForeignKey("VUser", on_delete="SET_NULL")` introspected to `on_delete='SET_NULL'`, emitted `ON DELETE UNSET`, parsed back as `'UNSET'` — states unequal, diff never converges.

Fixed at the producer rather than in `__eq__`: `_introspect_link_field` now stores `on_delete_to_surql(marker.on_delete)`, so `FieldState` holds the SurrealDB keyword the database will report back, the same way `field_type` already holds the SurrealDB type. Django's vocabulary stays the public API on `ForeignKey`; only the state normalizes.

Coverage: `test_field_state_stores_the_surrealdb_keyword`, `test_emitted_ddl_parses_back_to_an_equal_field_state` (PROTECT, asserting `parse_define_field(ddl) == field`), and the integration model now carries `reviewer: ForeignKey(..., on_delete="SET_NULL")` so `test_schema_diff_is_empty_after_define_table` exercises a non-CASCADE strategy end-to-end.

src/surreal_orm/migrations/operations.py
EulogySnowfall: Confirmed: `forwards()` emitted `REFERENCE ON DELETE CASCADE`, `backwards()` emitted `TYPE option<record<u>>;` with no clause at all.

`AlterField` gains `previous_reference` / `previous_on_delete`, and `backwards()` emits the clause from them. The related drift you flagged in the same block — `previous_flexible` / `previous_readonly` never forwarded from `state.py` — is fixed by the `from_field_states()` factory (see the state.py thread), which wires the whole previous state in one place.

`test_diff_forwards_every_previous_attribute` drives a real diff and asserts FLEXIBLE, READONLY and `REFERENCE ON DELETE UNSET` all survive into the rollback.

src/surreal_orm/fields/relation.py
EulogySnowfall: Confirmed: `AddField(..., on_delete="THEN").forwards()` emitted `REFERENCE ON DELETE THEN;`, which SurrealDB cannot parse.

Dropped `"THEN"` from the `OnDelete` literal rather than adding an expression parameter — nothing in the ORM needs `ON DELETE THEN <expression>` today, so adding a channel for it would be speculative. The alias now documents the omission and why. `test_then_is_not_an_accepted_strategy` guards it.

src/surreal_orm/migrations/operations.py
EulogySnowfall: Confirmed both halves: `AddField(field_type="int | string", nullable=True)` emitted a non-nullable column, and the `option<string | int>` round-trip re-dropped the wrap on every apply.

Took your first suggestion — the guard now mirrors `define_parser.py:182` exactly (`option<` prefix, `| null`, `none |` prefix) instead of testing for any pipe. Covered by `test_a_union_without_none_is_still_wrapped` and `test_a_union_carrying_none_is_left_alone`.

src/surreal_orm/migrations/introspector.py
EulogySnowfall: Confirmed: `ReferencesField["CamelWriter"]` on a model whose table is `CamelWriters` produced `array<record<camelwriter>>` while `ForeignKey` on the same target produced `record<CamelWriters>`.

Routed through `_resolve_target_table`, with one deliberate difference from the FK branch: the fallback keeps the marker's literal instead of degrading to untyped `record`. `ReferencesField` documents its parameter as the *table name*, so `ReferencesField["books"]` with no model registered for `books` must still emit `record<books>` — falling back to `record` there would be a regression for the documented usage.

The lowercasing had to stay for `get_references_info()`, which is tested to return it, so `_ReferencesMarker` now keeps the raw spelling in `.target` and resolution reads that. Covered by `test_references_field_resolves_a_model_name_to_its_table` and `test_references_field_keeps_an_unresolved_table_name`.

src/surreal_orm/migrations/introspector.py
EulogySnowfall: Confirmed: `author: ForeignKey("VUser") | None = None` introspected to `field_type='any', reference=False, nullable=False`.

Fixed rather than documented — `_get_relation_marker` now expands a top-level union and checks each member for `Annotated`. `test_foreign_key_wrapped_in_an_optional_union_is_still_a_record_link` covers it.

On the side note: the `Annotated` loop is redundant *today*, but this fix is what makes it load-bearing — for a union-wrapped marker, `FieldInfo.metadata` is empty and the annotation walk is the only path that finds it. So both loops stay, and now for a reason.

src/surreal_orm/migrations/state.py
EulogySnowfall: Agreed, and this was the most valuable finding — it is the defect class of #170 itself, and my patch was adding a fifth manual thread-through.

`AddField.from_field_state(table, state)` and `AlterField.from_field_states(table, current, target)` are now the only places a `FieldState` is expanded. Both diff branches and `define_table()` call them; the four hand-copied kwarg lists are gone. That also closes the live instance you named — `previous_flexible` / `previous_readonly` are wired now, along with the new `previous_reference` / `previous_on_delete`.

`define_table()`'s one special case (suppressing the encrypted VALUE clause on USER tables, since `define_access()` SIGNUP already hashes) is applied to the built operation rather than by rebuilding the kwarg list.

src/surreal_orm/fields/relation.py
EulogySnowfall: Correct — the truthy `if self.on_delete:` gate emits `REFERENCE ON DELETE IGNORE`, not a bare `REFERENCE`. Docstring corrected to describe what the code does: `IGNORE` is emitted as SurrealDB's own no-op keyword.

Kept the emission as-is for the reason you give: suppressing the clause would make the model side store `None` where the database reports `IGNORE`, reintroducing exactly the raw-string mismatch from the vocabulary thread.

EulogySnowfall:
Both body-level findings verified and handled in `9d83c28`; the eight inline threads have individual replies.

**`inspectdb` round-trip turns a scalar FK into `ReferencesField` — fixed.** Reproduced exactly as described: `record<users> REFERENCE ON DELETE CASCADE` generated `author: ReferencesField["users", "CASCADE"]`, which re-introspects as `array<record<users>>`. `_generate_field` now branches on arity — `array<record<T>>` / `set<record<T>>` (bare or `option<>`-wrapped) keeps `ReferencesField`, a scalar `record<T>` generates `ForeignKey(table, on_delete=...)`. Worth flagging: `tests/test_phase2_beta.py::test_generates_references_field_with_on_delete` asserted the defective mapping, so it is corrected rather than merely extended, and the `array<record<T>>` + `ON DELETE` case it had left uncovered is now tested. `test_generated_foreign_key_round_trips_without_a_type_change` asserts the model-side and database-side `FieldState` compare equal.

**One-time `REMOVE FIELD` for legacy M2M/Relation columns — accepted, documented.** The analysis is right and the behavior is intended: those columns were never read by the ORM, so dropping them loses nothing, but the statements are auto-generated and users should review rather than apply them blind. Added a dedicated section to the PR description covering both this and the `ON DELETE` change, flagged as needing a CHANGELOG / migration-guide entry. It is not in `CHANGELOG.md` here because this repo's release checklist puts docs in the release PR, not the fix PR — so this is a hand-off, and it needs to land there.

Verification after the fixes: `ruff check` clean, `mypy src/` clean on 66 files, unit 2004 passed / 26 skipped / 0 failed, integration against SurrealDB 3.2.4 454 passed / 1 skipped / 1 xfail (upstream #147) / 0 failed. Branch also merged up to `main` at `0.32.8`.

EulogySnowfall:
**Phase 8 — v2 backport verdict: partial, and it is separate work, not a cherry-pick.**

Checked `origin/v2` (0.21.4) rather than reasoning from `main`:

| Defect | On `v2`? |
| --- | --- |
| `ManyToMany` / `Relation` emitted as `any` columns | **Yes** — `_introspect_model` has the same unconditional `table_state.fields[field_name] = field_state` |
| Nullability dropped at the diff boundary | **Yes, and worse** — `AddField` / `AlterField` have no `nullable`, and `define_table()` does not even pre-wrap `option<>` there |
| `ForeignKey` falling through to `any` | **Yes**, for the typing half — record links do exist in SurrealDB 2.x |
| `REFERENCE` / `ON DELETE` clause | **No** — `FieldState` and the operations carry no `reference` / `on_delete` at all, and `ReferencesField` is absent from `fields/__init__.py` |

`REFERENCE` is a SurrealDB 3.0 feature, so the half of this PR that types record links and emits `REFERENCE ON DELETE` must not be carried to a 2.6.x branch. `v2` also lacks `on_delete_to_surql()` (that is #169, likewise not backported), so the normalization this PR depends on is not there either.

What genuinely applies to `v2` is the virtual-field exclusion and the nullability plumbing — real correctness bugs on the LTS line, but re-implemented against different code rather than cherry-picked. Not opening a backport PR unprompted; say the word if you want it as its own iteration.

Issue #55, opened 2026-02-15
Parameter binding fails for large nested dicts (saves as empty object)


EulogySnowfall:
## Description

When using `$param` parameter binding with `_raw_query()` or ORM `.save()` for large/deeply nested dict fields, SurrealDB receives an empty object `{}` instead of the actual data.

## Reproduction

```python
# This saves game_state as {} in the database
query = "UPDATE game_tables:abc SET game_state = $game_state"
await adapter._raw_query(query, {"game_state": large_nested_dict})

# Same with ORM .save() — the nested dict is stripped
table.game_state = {"scores": [100, 200], "hands": {"player1": [{"suit": "hearts", "value": "A"}]}}
await table.save()  # game_state saved as {}
```

## Current workaround

Embed JSON directly in the query string (bypasses parameter binding entirely):

```python
import json
json_str = json.dumps(state, default=str)
await adapter._raw_query(f"UPDATE game_tables:{id} SET game_state = {json_str}")
```

This works but bypasses the safety of parameterized queries. While the data is server-generated (not user input), it's not ideal.

## Root cause hypothesis

The issue likely occurs during CBOR serialization of the parameter dict — large or deeply nested structures may exceed some internal limit or trigger a serialization bug in the SDK's CBOR encoder.

## Affected patterns

This impacts any field that stores complex nested data:
- `game_state` (game tables)
- `game_state_before` / `game_state_after` (game actions)
- `archive_data` (game history)

Fields declared as `FLEXIBLE TYPE option<object>` work correctly when data is embedded directly in the query.

## Environment

- surrealdb-orm: 0.14.2
- SurrealDB: v2.6.0
- Python: 3.13

## Expected behavior

Parameter binding should correctly serialize and transmit large nested dicts to SurrealDB, matching the behavior of direct JSON embedding in queries.

PR #193, opened 2026-09-05
fix(orm): stop re.sub parsing inlined JSON as a replacement pattern

EulogySnowfall:
No issue — this surfaced from a six-month-old `git stash` on `main` ("unknown utils.py change - re.sub lambda fix", 2026-03-09). The buggy line had not moved since, so the fix still applied verbatim. Rather than dropping the stash, I verified the bug is live and turned it into a real change with tests.

### The bug

`inline_dict_variables()` passed the serialised JSON straight to `re.sub` as a **replacement string**. `re.sub` reads that as a mini-pattern: `\1` is a backreference, `\g<0>` a group reference, and an unknown escape is an error. JSON is full of both.

Reproduced against the current `main`:

| Value in a nested dict | Before | After |
| --- | --- | --- |
| `{"nom": "café"}` | **`re.PatternError: bad escape \u`** — the whole query aborts | `"caf\u00e9"`, correct |
| `{"path": "C:\temp\x"}` | `C:\temp\x` — one backslash lost per pair | `C:\\temp\\x`, correct |
| `{"v": "\g<0>"}` | silently corrupted | preserved |
| `{"v": "\1"}` | silently corrupted | preserved |

`json.dumps` emits `\uXXXX` for **every non-ASCII character**, so any accented text crashes the call outright; and it doubles every literal backslash, which `re.sub` then collapses — quietly changing the data that reaches the database.

Two things make this worse than the line count suggests:

- This is the inlining path added in **0.14.3** specifically to carry *complex nested data* past the CBOR variable-binding limits of #55 — precisely the data most likely to contain an accent or a path.
- The backslash cases fail **silently**. The crash is loud and gets noticed; the corruption does not.

### The fix

A **callable** replacement is exempt from that parsing — `re.sub` uses its return value as-is.

```python
query = re.sub(rf"\${re.escape(key)}\b", _literal_replacement(json_str), query)
```

`_literal_replacement()` takes the text as a parameter rather than closing over the loop variable. That is not cosmetic: the stashed version used a bare `lambda _: json_str`, which trips ruff **B023** (closure over a loop variable), and the obvious workaround — `lambda _, v=json_str: v` — then fails mypy with `Cannot infer type of lambda`. A named, annotated helper satisfies both and gives the *why* one place to live.

### Verification

| Command | Result |
| --- | --- |

Four regression cases added to `tests/test_issue_55_unit.py`, next to the inlining tests they belong with. Each parses the emitted JSON back and compares it to the input, which is the assertion that actually catches this — two of them failed with `PatternError` before the fix, the other two with `JSONDecodeError` from the collapsed backslashes.

> A note on the mypy gate while working on this: running the CLI tests first installs the `cli` extra, after which `mypy` reports 38 spurious `unused-ignore` errors in `cli/commands.py`. `CLAUDE.md` documents this; `uv sync --group lint --exact` restores a CI-equivalent environment. Worth running the CLI suite *after* the gates, not before.

### Release notes required

Docs are not touched here, per the release-checklist convention. The release PR needs:

- **Fixed** — `raw_query(inline_dicts=True)` crashed with `re.error` on accented text and silently dropped backslashes from strings such as Windows paths; astral characters (emoji) were then rejected by SurrealDB 3.x as invalid escape sequences. Inlined JSON is now substituted verbatim, with `ensure_ascii=False`.
- **Fixed** — live queries inlined their filter values through the same defective path, so a backslash in a `QuerySet.live()` filter silently matched nothing.
- **Fixed** — inlining two complex variables in one query could rewrite text inside the first one's JSON, with the result depending on dict order.

### Backport

The same line exists verbatim on `v2` (`utils.py:274`), and it is a correctness bug, so it qualifies. I have **not** opened that PR — say the word and I will.

---

## Review round

Every point was reproduced before changing anything. Three defects were real and are fixed; the release note above is corrected.

**1. Astral characters still failed, server-side.** `json.dumps` defaults to `ensure_ascii=True`, emitting a UTF-16 surrogate pair for anything outside the BMP, which SurrealDB 3.2.4 rejects:

```
RETURN {"x": {"s": "\ud83d\ude42"}};
Parse error: String contains invalid escape sequence, unicode escape character is not a valid unicode character.
```

BMP text (`caf\u00e9`) does decode, so my original "any non-ASCII content" was too broad — corrected in the release note. `ensure_ascii=False` sends the character raw. Verified end to end through `raw_query`: an emoji, a ZWJ family sequence and a Windows path all come back from a live 3.2.4 identical to the input. The point about the tests not seeing this is right and worth keeping: `json.loads` accepts surrogate pairs, so the new test asserts on the emitted text instead of round-tripping.

**2. The second site was live.** `LiveSelectStream._inline_params_static` reproduced exactly:

```python
_inline_params_static('WHERE p = $p', {'p': 'C:\\temp\\x'})  # -> WHERE p = 'C:\temp\x'
```

**3. Cross-key rewriting reproduced verbatim**, including the order dependence.

All three now go through **`surreal_sdk.utils.substitute_params()`** — in the SDK, since the dependency only runs one way. It substitutes verbatim and in a single pass, which removes `_literal_replacement` and with it the B023 / mypy-inference workarounds, and lets `live_select` stop sorting keys by length (greedy `\w+` already prevents `$_f1` matching inside `$_f10`).

**Minor, all applied:** the three round-trip tests are parametrized and assert via `startswith` + `removeprefix`; added cases for astral characters, two complex variables in one query, and an unknown reference left alone; `re.PatternError` → `re.error` (3.13-only name, project supports 3.12); the rationale now lives once, in the helper docstring, with the call sites keeping only what is specific to them.

**Not done, deliberately:** the two edge cases flagged as low severity — a complex key that never matches being dropped from the bindings without error, and `$key` inside a quoted string literal being rewritten. Neither is closed by the single pass on its own, and both are behaviour changes rather than fixes to what this PR touches. Say the word if you want them in scope.

**Backport:** confirmed still warranted, and the `live_select` line is identical on `v2`. `substitute_params` carries over as-is; `ensure_ascii=False` is a no-op there behaviourally (2.6.5 accepts surrogate pairs) but keeps the branches identical.

### DIFF
--- a/src/surreal_orm/utils.py
+++ b/src/surreal_orm/utils.py
@@ -227,6 +227,30 @@ def _extract_datetime_values(
     return value
 
 
+def _literal_replacement(literal: str) -> Callable[[re.Match[str]], str]:
+    """
+    Build a ``re.sub`` replacement that inserts *literal* verbatim.
+
+    ``re.sub`` reads a replacement **string** as a mini-pattern: ``\\1`` is a
+    backreference, ``\\g<0>`` a group reference, and an unknown escape is an error.
+    JSON is full of both — ``json.dumps`` emits ``\\uXXXX`` for every non-ASCII
+    character and doubles every literal backslash — so passing it directly raises
+    ``bad escape \\u`` on any accented text and silently collapses the backslashes
+    in a Windows path.
+
+    A callable replacement is exempt from that parsing: its return value is used
+    as-is. Taking the text as a parameter (rather than closing over a loop
+    variable) keeps the binding unambiguous.
+
+    Args:
+        literal: The exact text to substitute
+
+    Returns:
+        A replacement callable suitable for :func:`re.sub`
+    """
+    return lambda _match: literal
+
+
 def inline_dict_variables(
     query: str,
     variables: dict[str, Any],
@@ -271,7 +295,7 @@ def inline_dict_variables(
                 json_str = json_str.replace(f'"{marker}"', literal)
 
             # Replace $key with inline JSON (word-boundary to avoid partial matches)
-            query = re.sub(rf"\${re.escape(key)}\b", json_str, query)
+            query = re.sub(rf"\${re.escape(key)}\b", _literal_replacement(json_str), query)
         else:
             remaining[key] = value
     return query, remaining
--- a/src/surreal_orm/utils.py
+++ b/src/surreal_orm/utils.py
@@ -4,10 +4,10 @@ import json
 import logging
 import random
 import re
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from typing import Any, TypeVar
 
-from surreal_sdk.utils import substitute_params
+from surreal_sdk.utils import find_param_references, substitute_params
 
 logger = logging.getLogger(__name__)
 
@@ -199,6 +199,10 @@ class _SurrealJSONEncoder(json.JSONEncoder):
         return super().default(o)
 
 
+#: A ``__SURQL_DT_N__`` placeholder, including the JSON quotes around it.
+_DATETIME_MARKER = re.compile(r'"(__SURQL_DT_\d+__)"')
+
+
 def _extract_datetime_values(
     value: Any,
     markers: dict[str, str],
@@ -229,6 +233,16 @@ def _extract_datetime_values(
     return value
 
 
+def _expand_datetime_markers(json_str: str, markers: Mapping[str, str]) -> str:
+    """Swap each quoted ``__SURQL_DT_N__`` placeholder for its ``d"..."`` literal.
+
+    Takes *markers* as a parameter rather than closing over the caller's loop
+    variable, which ruff flags (B023) and which would be a real bug if the
+    callable outlived the iteration.
+    """
+    return _DATETIME_MARKER.sub(lambda match: markers[match.group(1)], json_str)
+
+
 def inline_dict_variables(
     query: str,
     variables: dict[str, Any],
@@ -253,11 +267,20 @@ def inline_dict_variables(
 
     Returns:
         ``(modified_query, remaining_variables)`` tuple.
+
+    Raises:
+        ValueError: If a referenced complex value cannot be serialized.
     """
+    # Which names the query actually uses. Without this a complex value whose
+    # reference is absent — or not a valid identifier, such as ``$my-var`` —
+    # would be serialized for nothing and then dropped from both the query and
+    # the returned bindings, silently.
+    referenced = find_param_references(query)
+
     inlined: dict[str, str] = {}
     remaining: dict[str, Any] = {}
     for key, value in variables.items():
-        if not _is_complex_value(value):
+        if key not in referenced or not _is_complex_value(value):
             remaining[key] = value
             continue
 
@@ -271,13 +294,19 @@ def inline_dict_variables(
             # them as UTF-16 surrogate pairs (\ud83d\ude42), which SurrealDB 3.x
             # rejects at parse time: "String contains invalid escape sequence".
             json_str = json.dumps(processed, cls=_SurrealJSONEncoder, ensure_ascii=False)
-        except (TypeError, ValueError) as e:
+            # ensure_ascii=False also lets a lone surrogate through json.dumps;
+            # encoding here keeps the failure inside the promised ValueError
+            # instead of surfacing as a UnicodeEncodeError in the CBOR encoder.
+            json_str.encode("utf-8")
+        except (TypeError, ValueError, UnicodeEncodeError) as e:
             raise ValueError(f"Failed to serialize variable '{key}' to JSON for inlining: {e}") from e
 
-        # Replace datetime marker strings (with JSON quotes) with
-        # unwrapped SurrealQL d"..." literals.
-        for marker, literal in dt_markers.items():
-            json_str = json_str.replace(f'"{marker}"', literal)
+        # Replace datetime marker strings (with JSON quotes) with unwrapped
+        # SurrealQL d"..." literals — one pass, for the same reason the query
+        # substitution is one pass: a str.replace per marker rescans the whole
+        # payload, which is quadratic in the number of datetimes.
+        if dt_markers:
+            json_str = _expand_datetime_markers(json_str, dt_markers)
 
         inlined[key] = json_str
 
--- a/src/surreal_orm/utils.py
+++ b/src/surreal_orm/utils.py
@@ -7,6 +7,8 @@ import re
 from collections.abc import Callable
 from typing import Any, TypeVar
 
+from surreal_sdk.utils import substitute_params
+
 logger = logging.getLogger(__name__)
 
 F = TypeVar("F", bound=Callable[..., Any])
@@ -227,30 +229,6 @@ def _extract_datetime_values(
     return value
 
 
-def _literal_replacement(literal: str) -> Callable[[re.Match[str]], str]:
-    """
-    Build a ``re.sub`` replacement that inserts *literal* verbatim.
-
-    ``re.sub`` reads a replacement **string** as a mini-pattern: ``\\1`` is a
-    backreference, ``\\g<0>`` a group reference, and an unknown escape is an error.
-    JSON is full of both — ``json.dumps`` emits ``\\uXXXX`` for every non-ASCII
-    character and doubles every literal backslash — so passing it directly raises
-    ``bad escape \\u`` on any accented text and silently collapses the backslashes
-    in a Windows path.
-
-    A callable replacement is exempt from that parsing: its return value is used
-    as-is. Taking the text as a parameter (rather than closing over a loop
-    variable) keeps the binding unambiguous.
-
-    Args:
-        literal: The exact text to substitute
-
-    Returns:
-        A replacement callable suitable for :func:`re.sub`
-    """
-    return lambda _match: literal
-
-
 def inline_dict_variables(
     query: str,
     variables: dict[str, Any],
@@ -276,29 +254,37 @@ def inline_dict_variables(
     Returns:
         ``(modified_query, remaining_variables)`` tuple.
     """
+    inlined: dict[str, str] = {}
     remaining: dict[str, Any] = {}
     for key, value in variables.items():
-        if _is_complex_value(value):
-            # Extract datetime objects as markers so they become d"..." literals
-            dt_markers: dict[str, str] = {}
-            counter = [0]
-            processed = _extract_datetime_values(value, dt_markers, counter)
-
-            try:
-                json_str = json.dumps(processed, cls=_SurrealJSONEncoder)
-            except (TypeError, ValueError) as e:
-                raise ValueError(f"Failed to serialize variable '{key}' to JSON for inlining: {e}") from e
-
-            # Replace datetime marker strings (with JSON quotes) with
-            # unwrapped SurrealQL d"..." literals.
-            for marker, literal in dt_markers.items():
-                json_str = json_str.replace(f'"{marker}"', literal)
-
-            # Replace $key with inline JSON (word-boundary to avoid partial matches)
-            query = re.sub(rf"\${re.escape(key)}\b", _literal_replacement(json_str), query)
-        else:
+        if not _is_complex_value(value):
             remaining[key] = value
-    return query, remaining
+            continue
+
+        # Extract datetime objects as markers so they become d"..." literals
+        dt_markers: dict[str, str] = {}
+        counter = [0]
+        processed = _extract_datetime_values(value, dt_markers, counter)
+
+        try:
+            # ensure_ascii=False keeps astral characters raw. The default emits
+            # them as UTF-16 surrogate pairs (\ud83d\ude42), which SurrealDB 3.x
+            # rejects at parse time: "String contains invalid escape sequence".
+            json_str = json.dumps(processed, cls=_SurrealJSONEncoder, ensure_ascii=False)
+        except (TypeError, ValueError) as e:
+            raise ValueError(f"Failed to serialize variable '{key}' to JSON for inlining: {e}") from e
+
+        # Replace datetime marker strings (with JSON quotes) with
+        # unwrapped SurrealQL d"..." literals.
+        for marker, literal in dt_markers.items():
+            json_str = json_str.replace(f'"{marker}"', literal)
+
+        inlined[key] = json_str
+
+    # One pass over the original query: substituting key by key would re-scan
+    # the JSON just inserted, so a "$b" inside a string value of $a would be
+    # replaced too, with the outcome depending on dict order.
+    return substitute_params(query, inlined), remaining
 
 
 def retry_on_conflict(
--- a/src/surreal_sdk/__init__.py
+++ b/src/surreal_sdk/__init__.py
@@ -74,6 +74,7 @@ from .types import (
     RecordsResponse,
     ResponseStatus,
 )
+from .utils import find_param_references, substitute_params
 
 __version__ = "0.33.2"
 __all__ = [
@@ -135,6 +136,9 @@ __all__ = [
     "TimeoutError",
     "TransactionError",
     "TransactionConflictError",
+    # Query-string helpers
+    "find_param_references",
+    "substitute_params",
 ]
 
 
--- a/src/surreal_sdk/streaming/live_select.py
+++ b/src/surreal_sdk/streaming/live_select.py
@@ -13,6 +13,7 @@ from typing import Any, Self
 
 from ..connection.websocket import WebSocketConnection
 from ..exceptions import LiveQueryError
+from ..utils import substitute_params
 
 _SAFE_TABLE_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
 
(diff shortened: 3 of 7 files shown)


EulogySnowfall:
## Review

The fix is correct for what it targets, and the four new tests do fail on `main`. Not ready to merge as is, though. Every point below was reproduced locally; the server-side ones were run against the `surrealdb/surrealdb:v3.2.4` container.

### Blocking

**1. Emoji still fails, now server-side.** `json.dumps` runs with the default `ensure_ascii=True`, so any character outside the BMP is emitted as a UTF-16 surrogate pair. SurrealDB 3.x dropped surrogate-pair support in string literals (it is listed in the 2.x → 3.x migration guide), and 3.2.4 rejects it:

```
RETURN {"x": {"s": "\ud83d\ude42"}}
Parse error: String contains invalid escape sequence, unicode escape character is not a valid unicode character.
```

BMP text (`caf\u00e9`) does decode, so the PR's `café` example holds, but "any non-ASCII content" does not. `ensure_ascii=False` at `utils.py:288` sends `🙂` raw and the server accepts it; control characters still go out as `\u001b`, which 3.x accepts. Please add an emoji case to the tests. Note the new tests cannot see this: they validate with `json.loads`, and surrogate pairs are valid JSON.

**2. The same bug is live at a second site.** `LiveSelectStream._inline_params_static` (`src/surreal_sdk/streaming/live_select.py:215`) also hands a string to `re.sub` as the replacement. `_format_value` doubles backslashes on purpose and the replacement parsing collapses them right back:

```python
_inline_params_static('WHERE p = $p', {'p': 'C:\\temp\\x'})  # -> WHERE p = 'C:\temp\x'
```

SurrealQL then reads `\t` as a tab and the live filter silently matches nothing. `QuerySet.live()` feeds user filter values into this path. The PR's diagnosis is applied to one of two sites. The SDK cannot import from the ORM, so if you want a shared helper it has to live in `surreal_sdk`.

### Should fold in, since the line is being touched

**3. Later keys rewrite text inside earlier inlined JSON.** The loop re-scans the whole query after each substitution, so a `$b` inside a string value of `$a` is replaced too, and the outcome depends on dict insertion order:

```python
inline_dict_variables('UPDATE t:1 SET a = $a, b = $b',
    {'a': {'x': {'s': 'cost $b here'}}, 'b': {'y': [1]}})
# -> UPDATE t:1 SET a = {"x": {"s": "cost {"y": [1]} here"}}, b = {"y": [1]}
```

Pre-existing, but a single pass fixes it and removes the need for `_literal_replacement` entirely: build `inlined = {key: json_str}` first, then one

```python
query = re.sub(r"\$(\w+)", lambda m: inlined.get(m.group(1), m.group(0)), query)
```

Greedy `\w+` keeps the existing guarantee that `$state` never matches `$state_backup`. It is also ~6x faster on large payloads (the query is copied once instead of once per key), and there is no loop variable to close over, so the B023 / mypy dance goes away with it.

### Minor

- If you keep the per-key shape, `json_str.join(re.split(pattern, query))` inserts the text verbatim with no callable, passes ruff B and mypy strict, and removes the 22-line factory. Either way the rationale is currently written three times (helper docstring, test-class docstring, PR body). Once is enough.
- The test docstrings name `re.PatternError`, which exists only on Python 3.13+. The project supports 3.12, where it is `re.error`.
- The three round-trip tests share one body; a single parametrized test would do. Asserting `new_query.startswith(prefix)` then parsing `new_query.removeprefix(prefix)` is stronger than slicing by a hard-coded length. None of the new tests cover two complex variables in one query, which is exactly where point 3 lives.
- **The proposed release note is inaccurate.** It says the fix covers "the complex-nested-data save path". The only caller of `inline_dict_variables` in the repo is `raw_query(inline_dicts=True)` (`model_base.py:1672`). The save path binds each field as a CBOR parameter through `_build_set_clause` and never inlines JSON, so it was never affected. Please drop that clause before it reaches the changelog.
- Two pre-existing edge cases that the single-pass rewrite closes naturally: a complex key that never matches (or is not a valid identifier) is silently dropped from the returned bindings with no error, whereas `_build_set_clause` validates with `_SAFE_IDENTIFIER_RE` and raises; and `$key` inside a quoted string literal in the query is rewritten, where parameter binding would have left it alone. Low severity given the trusted-input warning, noted for completeness.

### Checked and fine

ruff, mypy strict, all four new tests pass on the branch and fail on `main`. The B023 and "cannot infer type of lambda" claims in the PR body are both true. `re.sub` does not retain the callable, so no closure leak. Field access on an inlined object (`{...}.a.b`) is valid SurrealQL on 3.2.4, so the `\b` boundary before `.` is fine. Not touching docs or version strings matches the release checklist.

### Backport

Warranted, and the live_select site has the same line on `v2`. SurrealDB 2.6.5 still accepts surrogate pairs, so `v2` works without `ensure_ascii=False`, but carrying it keeps the two branches identical.

EulogySnowfall:
Review addressed in `cf3a316`. Every point reproduced before changing anything; CI green (12 checks).

**1. Astral characters — confirmed, fixed.** Reproduced against 3.2.4 verbatim:

```
RETURN {"x": {"s": "\ud83d\ude42"}};
Parse error: String contains invalid escape sequence, unicode escape character is not a valid unicode character.
```

`caf\u00e9` does decode, so my "any non-ASCII content" was too broad — corrected in the release note. `ensure_ascii=False` now sends the character raw, verified end to end: an emoji, a ZWJ family sequence and a Windows path go through `raw_query` into a live 3.2.4 and come back identical to the input. Your note about the tests being blind to this was the useful half — `json.loads` accepts surrogate pairs, so the new test asserts on the emitted text rather than round-tripping.

**2. Second site — confirmed, fixed.** `_inline_params_static('WHERE p = $p', {'p': 'C:\\temp\\x'})` gave `WHERE p = 'C:\temp\x'` exactly as you described.

**3. Cross-key rewriting — confirmed, fixed**, order dependence included.

All three now go through **`surreal_sdk.utils.substitute_params()`**, in the SDK as you pointed out it would have to be. Verbatim insertion plus a single pass over the original query. Taking the single-pass suggestion was the right call beyond correctness: it removes `_literal_replacement` and with it both workarounds it existed for, and `live_select` no longer needs to sort keys by length since the greedy `\w+` already prevents `$_f1` matching inside `$_f10`.

**Minor — all applied.** Round-trip tests parametrized, asserting via `startswith` + `removeprefix`; added astral characters, two complex variables in one query, and an unknown reference left alone; `re.PatternError` → `re.error` (3.13-only, project floor is 3.12); rationale reduced to one place, the helper docstring, with each call site keeping only what is specific to it.

**Release note — you were right and it mattered.** I checked: `inline_dict_variables` has exactly one caller, `raw_query` at `model_base.py:1672`. The save path binds through `_build_set_clause` and never inlines, so the clause I wrote would have put a false statement in the changelog. Dropped, and replaced with the two additional fixes this round brings.

**Two things I did not do**, flagged for your call rather than silently skipped: the complex key that never matches being dropped from the bindings without error, and `$key` inside a quoted string literal being rewritten. Neither is actually closed by the single pass — both would be deliberate behaviour changes beyond what this PR touches. Happy to fold them in if you want them here.

**Backport:** confirmed still warranted, and the `live_select` line is identical on `v2`. `substitute_params` carries over unchanged; `ensure_ascii=False` is behaviourally a no-op there (2.6.5 accepts surrogate pairs) but keeps the branches identical, which is worth more than the saving.

EulogySnowfall:
## Rebased on `main`, and the blast radius is wider than the title suggests

Rebased onto `33e02ed`+ (now `2ddd762`, 0.33.2) — conflict-free, since nothing merged this week touched `utils.py` or the SDK.

While confirming the defect still reproduces before rebasing, I found the common trigger is not an exotic value. `json.dumps` escapes **every non-ASCII character** as `\uXXXX`, and that JSON is handed to `re.sub` as a *replacement template*, where `\u` is an invalid escape. So on `main` today:

```python
inline_dict_variables("UPDATE t SET a = $v;", {"v": {"meta": {"k": 1}, "p": "café"}})
# PatternError: bad escape \u at position 28
```

Same for `🎉` and `北京`. `raw_query(inline_dicts=True)` therefore crashes on any accented text — which for most real data is "always", not "in an edge case".

Verified on this branch, all previously-failing inputs plus the group-reference forms that motivated the fix:

| input | before | after |
|---|---|---|
| `café` / `🎉` / `北京` | `PatternError` | ✅ emitted literally, not as `\u00e9` |
| `C:\\path` | ✅ | ✅ |
| `\1 et \g<0>` | `PatternError` | ✅ |
| `it's` | ✅ | ✅ |

Emitting the accented text literally rather than as `\uXXXX` is the better shape for SurrealQL too.

## Note: `main` is currently behind `v2` here

I fixed the same defect on the v2 line in #198 — the ForeignKey backport put a backslash-bearing value on that code path, so it surfaced there. `v2` accepts `café` and `🎉` today; `main` does not. This PR closes that gap in the right direction.

## Verification

Unit and integration suites green against 3.2.4, ruff clean, mypy strict clean both with and without `click` installed (that second environment matters since #197 put `click` in the dev group).

EulogySnowfall:
## Review, round 2

**Mergeable.** Every blocking point from the first round is fixed, and I confirmed each one end to end rather than from the diff alone.

- **Emoji**: `ensure_ascii=False` is in, with a test that asserts on the emitted text rather than a `json.loads` round trip, which is the right assertion. An inlined dict containing `Zoé 🙂 👨‍👩‍👧`, a Windows path, quotes and `\1` comes back from SurrealDB 3.2.4 equal to the input through `raw_query`.
- **Second site**: `_inline_params_static` now goes through the same helper, with three regression tests.
- **Single pass**: `substitute_params` lives in the SDK, dependency in the right direction, and the cross-key rewrite is gone.
- **Release note** no longer claims the save path was affected. `PatternError` is gone from the docstrings. The tests are parametrized with `startswith` plus `removeprefix`.

Gates: ruff, ruff format, mypy strict (68 files), the full unit suite, and the inlining and live-query integration tests against 3.2.4 all pass.

### Worth folding in, same function

1. **A complex key that is never consumed is silently dropped.** It goes into `inlined`, is removed from `remaining`, and if no `$key` matched it vanishes from both the query and the bindings. Two ways to hit it: the reference is absent from the query (pre-existing), or the key contains a non-`\w` character such as `my-var`, which the old `re.escape(key)` path did substitute and the new one leaves as a dangling `$my-var`:

   ```python
   inline_dict_variables('SET a = $my-var', {'my-var': {'x': {'y': 1}}})
   # main:    ('SET a = {"x": {"y": 1}}', {})
   # PR head: ('SET a = $my-var', {})
   ```

   Neither is a valid use, but the failure is now silent where it used to be visible. Collecting `set(_PARAM_REFERENCE.findall(query))` before the loop closes both at once: an unreferenced key is neither serialized (52 ms wasted on a large dict today) nor dropped, it stays in `remaining`.

2. **A lone surrogate escapes as the wrong error.** With `ensure_ascii=False`, a string like `"ab\udcff"` passes `json.dumps`, and the failure surfaces later as `UnicodeEncodeError` inside the CBOR encoder instead of the `ValueError("Failed to serialize variable ...")` the function promises. A `json_str.encode("utf-8")` inside the same `try` keeps the contract.

3. **`\w` is Unicode-aware; the old lookahead was ASCII.** `$pé` was substituted before and is left alone now. SurrealDB's identifier lexer is ASCII, so `re.compile(r"\$(\w+)", re.ASCII)` matches the server and restores the old boundary. Low likelihood, one flag.

4. **The datetime-marker loop is quadratic.** Each `d"..."` literal is a full `str.replace` over the whole JSON string. Measured on the PR head: 5,000 datetimes in an 0.8 MB payload take 1.6 s in that loop, against 21 ms with the same callable-regex technique this PR introduces (`re.sub(r'"(__SURQL_DT_\d+__)"', lambda m: dt_markers[m.group(1)], json_str)`). Pre-existing, but the PR re-indents those lines and its stated purpose is "one pass with a callable replacement".

### Follow-ups, separate PRs

**`HTTPTransaction.commit()` has the same class of bug and the new helper is a drop-in fix.** It namespaces parameters with `str.replace`, with no identifier boundary:

```
UPDATE users SET a = $a WHERE id = $auth.id      with vars {"a": ...}
-> UPDATE users SET a = $tx_0_a WHERE id = $tx_0_auth.id
```

Any statement binding a key that is a prefix of `auth`, `session`, `this`, `before`, `after`, `value` or `input` renames the built-in to an unbound variable, so the `WHERE` compares against `NONE`. That is every `save(tx=)` and `merge(tx=)` on an HTTP connection. `substitute_params(stmt.sql, {k: f"$tx_{i}_{k}" for k in stmt.vars})` produces the correct output (verified).

**On `main`, the live inliner is a workaround for a server the branch no longer supports.** I ran `LIVE SELECT * FROM lv WHERE tag = $tag AND n > $n` with bound `{"tag": "want", "n": 5}` over WebSocket and created three records. On 3.2.4 exactly the matching one fired; on 2.6.5 nothing fired, which is why `_format_value` / `_inline_params_static` exist. Since `main` requires 3.2+ as of 0.32.0, `LiveSelectStream.start()` and `_resubscribe_one` can pass `self.params` to `connection.query()` and the two renderers can go. That also removes the pre-existing renderer defects: a dict filter value is rendered as a Python repr (`filter(prefs={"theme": "dark"}).live()` emits `WHERE prefs = '{\'theme\': \'dark\'}'` and never fires) and `Decimal` as a quoted string. This does not affect the PR: the fix is correct, needed on `v2` where binding does not work, and `raw_query(inline_dicts=True)` has no binding alternative. But `substitute_params` would then have a single caller. Related: the original #55 symptom (nested dicts bound as `{}`) no longer reproduces on either server, and `test_sdk_query_large_dict_variable` asserts the bound path works, so `inline_dicts` itself may be deprecable. Worth an issue, not this PR.

### Minor

- The `_inline_params_static` docstring restates what `substitute_params` documents, and because it is not a raw string the `\t` it mentions renders as whitespace, so the sentence about the tab is unreadable in `help()`.
- `test_longer_key_first` still says `$_f10` should be replaced before `$_f1`; the sort it guarded is gone and the test now passes for a different reason. Rename or re-docstring, otherwise the next reader reintroduces the sort.
- `test_a_longer_key_is_not_matched_by_a_shorter_one` duplicates the existing `test_variable_word_boundary` in the same module.
- `substitute_params` has no direct test; the same three properties are tested twice through the two wrappers. One `tests/sdk/test_utils.py` would replace them.
- `substitute_params` is framed as the shared SDK primitive but is not re-exported from `surreal_sdk.__all__`.
- `if not replacements: return query` is a dead branch. The `label` parameter of the parametrized test is unused; `pytest.param(..., id=...)` does the same job.
- A `$name` inside a quoted literal in the caller's query is still rewritten, as before. Not a regression, but the deleted docstring was the only place that mentioned string literals, so a one-line caveat on the helper would help.

EulogySnowfall:
Round-2 review addressed in `a6082f4`. Every point reproduced against the branch before changing anything; the two follow-ups are filed rather than folded in.

## Worth folding in — all four were real

**1. An unreferenced complex key is silently dropped.** Both halves reproduced:

```python
inline_dict_variables("SET a = 1", {"v": {"x": {"y": 1}}})
# before: ("SET a = 1", {})            <- the value is simply gone
inline_dict_variables("SET a = $my-var", {"my-var": {"x": {"y": 1}}})
# before: ("SET a = $my-var", {})      <- and this one regressed against main
```

Your fix is the one I took: `find_param_references(query)` collects the names first, so an unreferenced key is neither serialized nor dropped — it stays in `remaining`. Both cases now return the binding untouched. The `my-var` half is the one that mattered: main substituted it, the single pass left it dangling, and nothing said so.

**2. Lone surrogate.** Confirmed — `"ab\udcff"` passed `json.dumps` and failed later as `UnicodeEncodeError` in the CBOR encoder. `json_str.encode("utf-8")` inside the same `try` restores the documented `ValueError`.

**3. `re.ASCII`.** Applied. One correction to the framing: `$pé` was substituted *before and after* — old code matched it via `re.escape(key)`, new code via Unicode `\w`. The real effect of the flag is different and better: with the reference set from point 1, an accented name is now left alone **and kept as a binding**, which is what the server would do with it.

**4. Quadratic datetime loop.** Confirmed, and worse than your measurement on this machine — your 5,000-datetime / 0.8 MB payload took **2.80 s**, against **0.023 s** with the callable regex. The mapping goes in as a parameter (`_expand_datetime_markers`) rather than a closure, since B023 fires here for the same reason it did on the original lambda.

## Minor — all applied

`_inline_params_static` is a raw string now and keeps only what is specific to it. `test_longer_key_first` → `test_a_shorter_key_does_not_match_inside_a_longer_reference`, with a docstring saying the greedy `\w+` is what holds and the sort must not come back. Dropped `test_a_longer_key_is_not_matched_by_a_shorter_one` as a duplicate of `test_variable_word_boundary`. New `tests/sdk/test_utils.py` covers `substitute_params` and `find_param_references` directly, including a test pinning the `re.error` that the helper exists to avoid. Both are re-exported from `surreal_sdk.__all__`. Removed the dead `if not replacements` branch and the unused `label` parameter (`pytest.param(..., id=...)` instead). The string-literal caveat is now a `.. note::` on the helper.

## Follow-ups — filed, not folded in

- **#204** — `HTTPTransaction.commit()`. Reproduced exactly as you wrote it: `$auth.id` becomes `$tx_0_auth.id` and the `WHERE` compares against `NONE`. `substitute_params` is a drop-in.
- **#205** — the live inliner as a 2.x workaround, with your 3.2.4-vs-2.6.5 measurement, the dict-repr and `Decimal` renderer defects, and the note that `inline_dicts` itself may be deprecable.

## Verification

ruff + ruff format clean, mypy strict clean (68 files, `uv sync --group lint --exact`), full unit suite exit 0, integration suite exit 0 against a clean `surrealdb/surrealdb:v3.2.4`.

End to end through `raw_query(inline_dicts=True)` on that server, a single payload carrying `Zoé 🙂 👨‍👩‍👧`, `C:\temp\x`, `\g<0>`, `\1`, `he said "hi"` and a `datetime` comes back equal to the input, with the datetime a real `datetime` — so the marker rewrite still produces a `d"..."` literal rather than a string.

PR #194, opened 2026-09-06
fix(migrations): make a table definition survive the round trip

EulogySnowfall:
First of three PRs from the review of what shipped in **0.33.1** unreviewed. Three defects, one cause: a `DEFINE TABLE` does not survive the trip to the server and back.

No issue — these came out of reviewing #185 after the fact.

### 1. Declared table permissions never reached the database

`CreateTable.forwards()` emitted `PERMISSIONS` as a **second** `DEFINE TABLE`, which SurrealDB rejects once the table exists. Reproduced on 3.2.4 *and* 2.6.5:

```
DEFINE TABLE t SCHEMAFULL;                                 -> OK
DEFINE TABLE t PERMISSIONS FOR select WHERE $auth.id = id; -> ERR "The table 't' already exists"
in the database: DEFINE TABLE t TYPE NORMAL SCHEMAFULL PERMISSIONS NONE
```

So a model declaring `permissions={"select": "$auth.id = id"}` got a table with **no** rule. It fails closed rather than open — `PERMISSIONS NONE` denies record users — so this is broken access control, not leaked data. And since #168 made the executor surface per-statement errors, the *first* migration of any such model now aborts with `MigrationStatementError`.

`PERMISSIONS` is a clause of `DEFINE TABLE`; it is emitted as one now.

### 2. Rolling back a table change dropped the table

The diff reuses `CreateTable` to redefine an existing table, and its `backwards()` was `REMOVE TABLE`. A migration that merely changed a permission had this rollback:

```
forwards() : DEFINE TABLE Article SCHEMAFULL;
             DEFINE TABLE Article PERMISSIONS FOR select WHERE true;
backwards(): REMOVE TABLE Article;      ← and reversible = True
```

`CreateTable` gains an **overwrite mode** — `DEFINE TABLE OVERWRITE`, which the plain form's rejection made necessary anyway — and carries the previous definition so `backwards()` restores it. `from_table_states()` wires both sides in one place; `_render()` is shared by both directions, so a rollback cannot drift from the definition it restores. A table the migration actually *created* still rolls back to `REMOVE TABLE`.

### 3. The permissions the server reports could not be parsed back

```
PERMISSIONS FOR select WHERE $auth.id = id, FOR create, update, delete NONE
```

`_parse_permissions` required a `WHERE`, so the `NONE` group was dropped entirely, and the condition kept the comma separating the groups (`'$auth.id = id,'`). Both shapes parse now.

**This one is load-bearing.** Without it the first fix would only make the mismatch *visible*: permissions would finally land, and every `makemigrations` would then propose a redundant redefinition forever. Fixing the emitter alone would have looked right in unit tests and failed in practice.

### Verification

End to end against SurrealDB 3.2.4:

```
DDL émis : DEFINE TABLE a_guarded SCHEMAFULL PERMISSIONS FOR select WHERE $auth.id = id;
en base  : ... PERMISSIONS FOR select WHERE $auth.id = id, FOR create, update, delete NONE
diff après application : aucune opération
modif  forwards : DEFINE TABLE OVERWRITE a_guarded SCHEMAFULL PERMISSIONS FOR select WHERE true;
modif  backwards: DEFINE TABLE OVERWRITE a_guarded SCHEMAFULL PERMISSIONS FOR select WHERE $auth.id = id;
forwards accepté par le serveur : ['OK']
```

| Command | Result |
| --- | --- |

New tests: `tests/test_createtable_roundtrip_unit.py` (11 cases across the three pieces) and `tests/test_createtable_roundtrip_integration.py` (4 against a live server, including one that writes a row, applies the redefinition, rolls it back, and asserts the row is still there).

### Release notes required

- **Fixed** — table permissions declared in `model_config` never reached the database: the clause was emitted as a second `DEFINE TABLE`, which SurrealDB rejects for an existing table. Any model using `permissions={...}` was created without its access rules, and since 0.32.6 its first migration aborted.
- **Fixed** — a migration that changed a table definition rolled back with `REMOVE TABLE`, dropping the table and all its rows while reporting itself reversible.
- **Fixed** — the `PERMISSIONS` shape `INFO FOR DB` reports could not be parsed back, so a permissioned table diffed as changed on every run.

### Still to come

Two more PRs from the same review: **B** — `AlterField.backwards()` drops `VALUE` / `encrypted`, so rolling back an alter on an `Encrypted` password column leaves it storing plaintext (affects `main` and `v2`). **C** — `makemigrations` emits `DropTable` for every unmodelled table, including `_surreal_orm_migrations`. Then the same three on `v2`, plus the gaps in the #191 backport.

### DIFF
--- a/src/surreal_orm/migrations/__init__.py
+++ b/src/surreal_orm/migrations/__init__.py
@@ -31,6 +31,7 @@ from .model_generator import ModelCodeGenerator
 from .operations import (
     AddField,
     AlterField,
+    AlterTable,
     CreateIndex,
     CreateTable,
     DataMigration,
@@ -58,6 +59,7 @@ __all__ = [
     # Migrations
     "Migration",
     "Operation",
+    "AlterTable",
     "CreateTable",
     "DropTable",
     "AddField",
--- a/src/surreal_orm/migrations/db_introspector.py
+++ b/src/surreal_orm/migrations/db_introspector.py
@@ -195,6 +195,7 @@ class DatabaseIntrospector:
             view_query=table_props.get("view_query"),
             relation_in=table_props.get("relation_in"),
             relation_out=table_props.get("relation_out"),
+            comment=table_props.get("comment"),
             enforced=table_props.get("enforced", False),
         )
 
--- a/src/surreal_orm/migrations/define_parser.py
+++ b/src/surreal_orm/migrations/define_parser.py
@@ -27,6 +27,10 @@ _FIELD_KEYWORDS = [
     "TYPE",
 ]
 
+# The four actions a table PERMISSIONS clause can govern, in the order
+# SurrealDB reports them.
+_PERMISSION_ACTIONS = ("select", "create", "update", "delete")
+
 # Keywords that delimit clauses in a DEFINE TABLE statement.
 _TABLE_KEYWORDS = [
     "SCHEMAFULL",
@@ -127,8 +131,12 @@ def _consume_until_keyword(text: str, keywords: list[str]) -> tuple[str, str]:
         elif ch == ">":
             depth_angle -= 1
 
-        # Only check for keyword boundaries at depth 0
-        if depth_paren == 0 and depth_angle == 0:
+        # Only check for keyword boundaries at depth 0, and only where a clause
+        # could actually start: after whitespace. Without the left-hand check a
+        # keyword appearing inside an expression ended the clause — `PERMISSIONS
+        # FOR select WHERE meta.type != NONE` was cut at `type`, leaving the
+        # condition as `meta.` and the table type as `!= none`.
+        if depth_paren == 0 and depth_angle == 0 and (i == 0 or text[i - 1] in (" ", "\t", "\n")):
             upper_rest = text[i:].upper()
             for kw in keywords:
                 if upper_rest.startswith(kw):
@@ -333,6 +341,15 @@ def parse_define_table(statement: str) -> dict[str, Any]:
     # Permissions
     permissions = _parse_permissions(clauses.get("PERMISSIONS"))
 
+    # Comment — the server echoes it back single-quoted, with '' for a quote
+    comment: str | None = None
+    raw_comment = clauses.get("COMMENT")
+    if raw_comment:
+        raw_comment = raw_comment.strip()
+        if raw_comment.startswith("'") and raw_comment.endswith("'") and len(raw_comment) >= 2:
+            raw_comment = raw_comment[1:-1]
+        comment = raw_comment.replace("''", "'")
+
     # Materialized view (AS SELECT ... or AS (SELECT ...))
     # We re-extract from the raw body because _extract_clauses may split on
     # ``AS`` tokens inside the view query itself (e.g. ``count() AS total``).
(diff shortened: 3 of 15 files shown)


EulogySnowfall:
## Review

The core of the change holds: `OVERWRITE` preserves fields, indexes, events and rows (verified on 3.2.4), and a single `DEFINE TABLE ... PERMISSIONS` statement is what makes the clause land. But the parser change opens a permissions hole, and the new diff path has three defects that did not exist before. Every point below was reproduced on the PR branch; the server-side ones against `surrealdb/surrealdb:v3.2.4`.

### Blocking

**1. `FULL` is rendered as a silent deny.** The new `(FULL|NONE)` group puts the literal string `"FULL"` into the permissions dict, and `_render` wraps every value in `WHERE`. SurrealDB accepts `FOR select WHERE FULL`, evaluates `FULL` as a field reference, gets `NONE`, and denies:

```
DEFINE TABLE pw SCHEMAFULL PERMISSIONS FOR select WHERE FULL;   -- status OK
RETURN !!FULL;                                                  -- false
```

Two reachable paths in this PR: `inspectdb` on a table defined with `FOR select FULL` now generates `permissions={'select': 'FULL', 'create': 'NONE', ...}`, and `from_table_states` copies the introspected dict into `previous_permissions`, so the rollback of any overwrite on such a table emits `WHERE FULL`. A world-readable table becomes unreadable with no error. Before this PR the group was dropped so neither path could produce it. `_render` needs to emit `FOR x FULL` / `FOR x NONE` for the keyword values, and `_effective_permissions` needs a story for `FULL` (today it strips only `NONE`, so a DB-side per-action `FULL` compares unequal to every model and the diff never converges).

**2. `TYPE USER` reaches the DDL.** `from_table_states` forwards `target.table_type` verbatim. The `TableType.USER` model from the docs produces:

```
DEFINE TABLE OVERWRITE acct TYPE USER SCHEMAFULL PERMISSIONS FOR select WHERE $auth.id = id;
Parse error: Unexpected token `USER`, expected `NORMAL`, `RELATION`, or `ANY`
```

`define_table()` sanitises USER/STREAM/HASH to `None` (model_base.py ~617); the diff path does not. Any permission, changefeed or schema-mode change on a USER table is unmigratable through `makemigrations`.

**3. An overwrite with no `previous_*` is "reversible" to nothing.** `CreateTable(name='c', overwrite=True, permissions={...})` reports `reversible=True` and `backwards()` emits `DEFINE TABLE OVERWRITE c;`. Verified: that resets the table to `TYPE ANY SCHEMALESS PERMISSIONS NONE`. `overwrite` is a public field and the unit test documents building it by hand. `AlterField.__post_init__` already encodes `reversible = previous_type is not None`; the table operation needs the same gate.

**4. A view model is silently replaced by a plain table.** `ModelIntrospector` never populates `TableState.view_query`, so a model with `view_query` always diffs (schema_mode, permissions) and the overwrite it emits carries no `AS` clause. Verified: `DEFINE TABLE OVERWRITE v SCHEMAFULL` turns `DEFINE TABLE v ... AS SELECT ...` into `TYPE NORMAL SCHEMAFULL`. Before this PR the plain `DEFINE TABLE` was rejected with "already exists" and the executor raised.

### Should fold in

- `previous_permissions=current.permissions or None` keeps the server's four default `NONE` entries, so every generated migration file and rollback carries `FOR create WHERE NONE FOR update WHERE NONE FOR delete WHERE NONE`. Run both sides through the existing `_effective_permissions()`; it was added in 0.33.1 for exactly this.
- `describe()` still returns `Create table X` for an overwrite, so `makemigrations`, `schemadiff` and executor failure messages label a redefinition as a creation. Worth considering a dedicated `AlterTable(Operation)` with the `previous_*` fields instead of a flag on `CreateTable`: it fixes `describe()`, the `reversible` rule and the `CreateTable`/`DropTable` symmetry in one move, and mirrors the `AddField`/`AlterField` split the codebase already has.
- The group split is not quote-aware: `FOR select WHERE label = 'FOR sale', FOR create, update, delete NONE` parses to `{'select': "label = '"}`, the diff never converges, and the rollback is a parse error. The module already has `_consume_until_keyword` for this. Pre-existing and same class: `_extract_clauses` cuts the PERMISSIONS clause at any `type`, `as`, `comment` or `drop` token inside a condition.
- `COMMENT` has no `previous_comment` and `TableState` does not carry it, so an overwrite drops it in both directions.
- `define_table()` still emits a non-OVERWRITE `DEFINE TABLE` and never checks per-statement status, so on an existing table a changed `PERMISSIONS` is rejected with "already exists" and nothing is raised. The PR title says permissions now reach the database; that is only true for the first call on a fresh table, which is what the integration test's `REMOVE TABLE` fixture guarantees.

### Minor

- `if rules:` under `if permissions:` is dead; a non-empty dict always joins to a non-empty string.
- `.strip().rstrip(",").strip()` is unreachable: the lookahead already stops group 2 before the comma. The comment above it says the opposite.
- `_render` takes nine positionals that `forwards()`, `backwards()` and `from_table_states()` each spell out again. One tuple of attribute names and `**{f: getattr(self, f)}` / `getattr(self, f"previous_{f}")` collapses the copies; that is also the shape that would have caught `comment`.
- The "Tables to create" branch in `state.py` and `define_table()` are two more copies of the same nine kwargs; a `CreateTable.from_table_state()` beside `from_table_states()` would give the field-style pairing.
- `[\w\s,]+?` in the new regex is quadratic on a clause of repeated `FOR` tokens (80 KB takes 26 s). Only `INFO FOR DB` output reaches it, so not blocking; splitting on `\bFOR\b` first removes it.

### Checked and fine

ruff, mypy strict, the new unit and integration tests pass on the branch. Not touching docs or version strings matches the release checklist.

### Backport

#199 is a faithful port and its tests pass on 2.6.5, but every point above reproduces identically there (I re-ran `FULL`, the quoted-`FOR` truncation, `TYPE USER`, `WHERE NONE`, the bare-OVERWRITE reset and the view case against 2.6.5). Fix here first, then port, so the two branches do not diverge on a permissions bug.

EulogySnowfall:

All four blocking points reproduced against `surrealdb:v3.2.4` before anything was changed, and each fix verified against the same server. Pushed in `4d4fd53`.

### Blocking

**1. `FULL` rendered as a silent deny.** Confirmed exactly as described: the parser produced `{'select': 'FULL', …}`, `_render` emitted `FOR select WHERE FULL`, and the server stored it verbatim while `RETURN !!FULL` is `false`. `FULL` and `NONE` are keywords now — `FOR select FULL` — which is what the server itself reports back for a table defined that way.

On the `_effective_permissions` half: it strips `NONE` and keeps `FULL`, and bare `PERMISSIONS FULL` now expands to all four actions the way bare `NONE` already did. Without that the two spellings of the same thing compared unequal.

**2. `TYPE USER` reached the DDL.** Confirmed — `Parse error: Unexpected token \`USER\`, expected \`NORMAL\`, \`RELATION\`, or \`ANY\``. Sanitising moved into the renderer rather than being repeated at `from_table_states`, so no path can bypass it the way the diff path did.

**3. Bare overwrite "reversible" to nothing.** Confirmed, and worse than reported: I checked what the reset actually does. `DEFINE TABLE OVERWRITE c;` turned `SCHEMAFULL PERMISSIONS FOR select FULL` into `TYPE ANY SCHEMALESS PERMISSIONS NONE`. Rows survive; the definition does not.

**4. View silently replaced by a plain table.** Confirmed. `ModelIntrospector` filled neither `view_query` nor the relation clauses, and the overwrite dropped the `AS`. It fills them now, plus the comment.

### On the `AlterTable` suggestion — taken

You floated it as worth considering; it is the right call and it is what landed. Redefinition is `AlterTable`, which fixes `describe()` (a redefinition was labelled "Create table"), the `reversible` rule, and the `CreateTable`/`DropTable` symmetry in one move. `backwards()` raises rather than emitting the reset, since returning a statement that destroys the definition is worse than failing.

Both operations and the renderer are driven off one `TABLE_DEFINITION_FIELDS` tuple, which is what makes the `COMMENT` omission you spotted structurally impossible rather than fixed once.

### Folded in

- `previous_permissions` runs through `_effective_permissions`.
- The permissions split is quote-aware and no longer quadratic. `'FOR sale'` truncated at the quote exactly as you showed; the 80 KB clause went from 26 s to 0.02 s.
- Your pre-existing `_extract_clauses` note turned out to bite harder than described: `PERMISSIONS FOR select WHERE meta.type != NONE` left the condition as `meta.` **and** the table type as `!= none`. `_consume_until_keyword` requires a left-hand boundary now.
- `COMMENT` carried in both directions, with `TableState.comment` and a `comment` config key so the round trip closes.
- `define_table()` emits the OVERWRITE form and checks per-statement status.
- The nine-kwarg copies collapsed into `from_table_state()` / `from_table_states()`.
- The dead `if rules:`, the unreachable `.strip().rstrip(",")` and its inverted comment are gone.

### Two things your review did not cover

- A `relation_out=["a", "b"]` list — the documented config form — rendered as `OUT ['a', 'b']`. Pipe-separated now.
- Making `define_table()` check statement status revealed it was never idempotent: the second call failed with `The field 'x' already exists` and the error was swallowed. `AddField` gained an opt-in `overwrite` for that path; the migration default stays the plain form, since a migration adding an existing field *should* fail.

`test_relation_type.py::test_user_type_emitted` asserted the invalid DDL, so it is inverted with the server's parse error recorded.

### Verification

1998 unit tests and 465 integration tests green against 3.2.4, mypy strict clean, ruff clean. New integration tests pin the four blocking cases server-side.

The v2 port is updated in #199, re-verified on 2.6.5 rather than assumed.

Issue #195, opened 2026-09-06
Field-level PERMISSIONS are not modelled, so DEFINE FIELD OVERWRITE wipes them


EulogySnowfall:
Split out while fixing the `AlterField` rollback.

Field-level `PERMISSIONS` are not represented anywhere in the migration path:

- `FieldState` has no `permissions` attribute (only `TableState` does).
- `parse_define_field()` does not extract the clause.
- `AddField` / `AlterField` never emit it.

On their own those are just a missing feature. What makes it a defect is that `AlterField` emits **`DEFINE FIELD OVERWRITE`** (since #162), which replaces the entire definition. So any field permission set outside the ORM — by hand, by another tool, by a `RawSQL` migration — is silently dropped the next time the ORM alters that field.

Unlike the `VALUE` clause, this is not only a rollback problem: `forwards()` wipes it too, so it cannot be fixed by adding a `previous_*` slot. It needs the clause modelled end to end — parsed into `FieldState`, compared by the diff, emitted by both operations.

## Why it was not folded into the `VALUE` fix

Different shape and different blast radius. `VALUE` is lost only on rollback and has a concrete security consequence (an `Encrypted` column stops hashing), so it was worth fixing on its own. This one is a modelling gap that changes `FieldState` equality — it will make previously-equal states differ, which is exactly the kind of change that deserves its own review and its own release note.

## Suggested shape

- `FieldState.permissions: dict[str, str]` with `__eq__` participation.
- `parse_define_field()` reads the clause via the existing `_parse_permissions()` (which now handles both the `WHERE` and the grouped `NONE`/`FULL` shapes).
- `AddField` / `AlterField` emit it, and `AlterField` carries `previous_permissions`.
- Model-side: decide whether to expose it (a `Field(json_schema_extra=...)` marker, or config) or to treat DB-side permissions as read-only and simply preserve them.

That last decision is the real design question — preserving what the database has is enough to stop the destruction, and is much smaller than giving models a way to declare field permissions.

Issue #162, opened 2026-08-19
ForeignKey and migration improvements


leosilberg:
Hi 

Thanks for your work on this  library, it's the best one out there for python Surrealdb. I came across a few bugs while trying it out, ai summarized

```
  1. AlterField (migrations/operations.py)

  - forwards() must emit DEFINE FIELD OVERWRITE, not plain DEFINE FIELD. The code comments "DEFINE FIELD is idempotent
    — it creates or updates," but on SurrealDB v3.2.4 a plain DEFINE over an existing field is silently ignored — no
    error, no change — so every AlterField ever applied was a no-op that reported success. OVERWRITE is
    create-or-redefine, so it's also still correct on a missing field. (This is what migrations/orm_compat.py patches
    in our repo.)
  - backwards() has the same bug — a rollback re-defining the previous type is also a silent no-op.
  - The executor compounds it by swallowing per-statement errors: client.query() raises only on RPC-level failure, so
    a statement-level ERR inside a multi-statement operation (we hit this with a DataMigration) passes as success. The
    migrate loop should check each statement's status.

  2. ForeignKey record handling (model_base.py, query_set.py)

  The read path converts (_preprocess_db_record turns RecordId → "table:id" string) but nothing converts the other
  way, so FK fields only work against string columns — defeating the whole record<>/REFERENCE feature:

  - Writes: save()/create/merge send model_dump() output, so an FK holds a plain string that a record<> column
    rejects. The ORM knows which fields are FKs (the markers are right there); _execute_save should wrap "table:id"
    strings in RecordId for FK fields. That's exactly what we added in our base model — it belongs upstream.
  - Filters: .filter(fk="table:id") binds a CBOR string, and SurrealDB never equates a string with a record — silently
    empty results, the worst failure mode. The QuerySet has the model class; it should coerce filter values on FK
    fields to RecordId.
  - on_delete vocabulary: ForeignKey accepts Django's CASCADE | SET_NULL | PROTECT, but SurrealDB's clause takes
    CASCADE | UNSET | REJECT | IGNORE | THEN. The emitters just uppercase and pass through, so SET_NULL/PROTECT would
    generate invalid DDL. Needs the mapping (SET_NULL→UNSET, PROTECT→REJECT) or aligned literals.

  3. Migration typing (migrations/introspector.py, state.py)

  - ModelIntrospector._introspect_field never looks at relation markers — it doesn't even unwrap Annotated, so every
[...]

PR #196, opened 2026-09-06
fix(migrations): restore the VALUE clause when rolling back an AlterField

EulogySnowfall:
Second of three PRs from the review of what shipped unreviewed. Depends on nothing; #194 is independent.

### The defect

`AlterField.backwards()` re-emits `DEFINE FIELD OVERWRITE`, which replaces the **whole** definition — so every clause without a `previous_*` slot is silently dropped. Two were missing, and one of them matters:

```
AlterField.from_field_states("users", encrypted_password, plain_password)

forwards()  -> DEFINE FIELD OVERWRITE password ON users TYPE string;
backwards() -> DEFINE FIELD OVERWRITE password ON users TYPE string;
```

Rolling that back drops `VALUE crypto::argon2::generate($value)`. The column **stops hashing**, and every write after the rollback stores the password in plaintext. `Computed` fields lose their expression the same way.

**It has been live on `main` since 0.32.6.** The gap existed before, but a rollback was then a silent no-op, so nothing happened. #162's `OVERWRITE` fix — correct in itself — is what turned it into an active overwrite. That is worth noting for its own sake: fixing the no-op made a latent defect destructive, and nothing in the suite noticed.

`v2` has the same gap and will get the same fix in its own PR.

### The fix

`previous_value` and `previous_encrypted`, wired by `from_field_states()` — the single place a `FieldState` becomes operation arguments — and emitted by `backwards()` in the same position and with the same precedence as `forwards()` (`encrypted` wins over a plain `value`).

### Verification

Against a live SurrealDB 3.2.4: apply `forwards()`, then `backwards()`, then write to the column — the stored value comes back as an `$argon2` hash rather than the plaintext.

A second integration test asserts that `forwards()` **alone** really does store plaintext. Without it the first test could pass for the wrong reason — if `forwards()` happened to keep the clause, nothing would be proven about the rollback.

| Command | Result |
| --- | --- |

New: `tests/test_alterfield_rollback_unit.py` (7 cases — both clause kinds, precedence, no clause invented where there was none, position before `DEFAULT`, and a real diff producing a restoring rollback) and `tests/test_alterfield_rollback_integration.py` (2, described above).

### Deliberately out of scope — #195

Field-level `PERMISSIONS` are not modelled anywhere: not in `FieldState`, not parsed, not emitted. `OVERWRITE` therefore wipes any set outside the ORM. That is a different shape of problem — it affects `forwards()` as much as `backwards()`, so no `previous_*` slot can fix it, and closing it changes `FieldState` equality. Filed as #195 with a suggested design.

### Release notes required

- **Fixed** — rolling back a migration that altered an `Encrypted` column dropped its `VALUE crypto::argon2::generate($value)` clause, so the column stopped hashing and later writes stored passwords in plaintext. `Computed` fields lost their expression the same way. Live since 0.32.6, when `AlterField` began emitting `DEFINE FIELD OVERWRITE`.

### Still to come

**C** — `makemigrations` emits `DropTable` for every unmodelled table, including `_surreal_orm_migrations`. Then A, B and C on `v2`, plus the gaps in the #191 backport.

### DIFF
--- a/src/surreal_orm/migrations/define_parser.py
+++ b/src/surreal_orm/migrations/define_parser.py
@@ -211,6 +211,15 @@ def parse_define_field(statement: str) -> FieldState:
     # Parse ASSERT clause
     assertion = clauses.get("ASSERT") or None
 
+    # Parse COMMENT clause — echoed back single-quoted, with '' for a quote
+    comment: str | None = None
+    raw_comment = clauses.get("COMMENT")
+    if raw_comment:
+        raw_comment = raw_comment.strip()
+        if raw_comment.startswith("'") and raw_comment.endswith("'") and len(raw_comment) >= 2:
+            raw_comment = raw_comment[1:-1]
+        comment = raw_comment.replace("''", "'")
+
     # Parse REFERENCE clause (SurrealDB 3.0)
     reference = "REFERENCE" in clauses
     on_delete: str | None = None
@@ -233,6 +242,7 @@ def parse_define_field(statement: str) -> FieldState:
         value=value_expr,
         reference=reference,
         on_delete=on_delete,
+        comment=comment,
     )
 
 
--- a/src/surreal_orm/migrations/operations.py
+++ b/src/surreal_orm/migrations/operations.py
@@ -715,6 +715,8 @@ class AlterField(Operation):
     previous_nullable: bool = False
     previous_reference: bool = False
     previous_on_delete: str | None = None
+    previous_value: str | None = None
+    previous_encrypted: bool = False
 
     @classmethod
     def from_field_states(cls, table: str, current: "FieldState", target: "FieldState") -> "AlterField":
@@ -755,6 +757,8 @@ class AlterField(Operation):
             previous_readonly=current.readonly,
             previous_reference=current.reference,
             previous_on_delete=current.on_delete,
+            previous_value=current.value,
+            previous_encrypted=current.encrypted,
         )
 
     def __post_init__(self) -> None:
@@ -826,6 +830,14 @@ class AlterField(Operation):
             parts.append("FLEXIBLE")
 
         parts.append(f"TYPE {normalized_prev_type}")
+
+        # OVERWRITE replaces the whole definition, so a clause with no
+        # previous_* slot is dropped. Losing VALUE is the dangerous one: an
+        # Encrypted column stops hashing and stores plaintext from then on.
+        if self.previous_encrypted:
+            parts.append("VALUE crypto::argon2::generate($value)")
+        elif self.previous_value:
+            parts.append(f"VALUE {self.previous_value}")
 
         if self.previous_default is not None:
             if isinstance(self.previous_default, str):
--- a/src/surreal_orm/migrations/operations.py
+++ b/src/surreal_orm/migrations/operations.py
@@ -90,6 +90,145 @@ def _apply_nullable(field_type: str, nullable: bool) -> str:
     return f"option<{field_type}>"
 
 
+#: The VALUE expression that makes a column an ``Encrypted`` one.
+ARGON2_VALUE = "crypto::argon2::generate($value)"
+
+#: The attributes a ``DEFINE FIELD`` statement carries, in clause order.
+#: ``AddField.forwards``, ``AlterField.forwards`` and ``AlterField.backwards``
+#: are all driven off this one tuple. They used to be three hand-written
+#: renderers, and every clause added to one of them was eventually forgotten in
+#: another — that is #170 (REFERENCE), #195 (VALUE) and the DEFAULT divergence
+#: below, all the same omission.
+FIELD_DEFINITION_FIELDS = (
+    "field_type",
+    "nullable",
+    "flexible",
+    "value",
+    "default",
+    "assertion",
+    "readonly",
+    "reference",
+    "on_delete",
+    "comment",
+)
+
+
+def _previous_attr(name: str) -> str:
+    """
+    Name the ``previous_*`` slot holding *name*'s pre-change value.
+
+    All of them follow the ``previous_<field>`` pattern except the type, which
+    shipped as ``previous_type``. Generated migration files on disk spell it
+    that way, so it stays.
+
+    Args:
+        name: A field-definition attribute name
+
+    Returns:
+        The matching previous slot's attribute name
+    """
+    return "previous_type" if name == "field_type" else f"previous_{name}"
+
+
+def _render_default(default: Any) -> str:
+    """
+    Render a DEFAULT value as SurrealQL.
+
+    A bare Python ``repr`` is wrong three ways: ``True`` is not ``true``, a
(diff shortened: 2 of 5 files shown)


EulogySnowfall:
## Review

Correct for what it targets, and the integration test's plaintext-password assertion is the right behavioural anchor. Worth completing before merging, because the block directly under the new code has the same defect.

### Should fix before merge

**`backwards()` renders `DEFAULT` differently from `forwards()`.** It single-quotes every string with no function detection, no quote escaping, and Python `str()` for booleans. Reproduced on the branch:

```python
cur = FieldState(name="created", field_type="datetime", nullable=False, default="time::now()")
AlterField.from_field_states("t", cur, tgt).backwards()
# DEFINE FIELD OVERWRITE created ON t TYPE datetime DEFAULT 'time::now()';   (forwards emits DEFAULT time::now())
FieldState(name="flag", field_type="bool", default=True) -> DEFAULT True      (forwards emits true)
default="it's"                                             -> DEFAULT 'it's'  (parse error, rollback fails)
```

This is reachable from a generated migration: `makemigrations` reads the live schema by default since 0.33.1 and `_parse_default_value` returns `time::now()` as a bare string. Same class as the `VALUE` bug this PR fixes, one clause down.

### Structural

- `previous_encrypted` is a derived slot. Both producers treat `encrypted` as "VALUE is `crypto::argon2::generate($value)`" and never set both `encrypted=True` and `value`, so the precedence branch and its test encode a state no diff can emit. `from_field_states` can set `previous_value` to the argon2 expression when `current.encrypted` and `backwards()` keeps one `if self.previous_value:`.
- This is the third hand-rendered `DEFINE FIELD` body (`AddField.forwards`, `AlterField.forwards`, `AlterField.backwards`), and the fourth copy of the argon2 literal counting `define_parser`. One `_render_define_field(...)` called by all three, with `backwards()` passing `getattr(self, f"previous_{k}")`, means a clause cannot be forgotten on one side. #170, this PR, #195 and `COMMENT` (which `AddField` renders and `AlterField` has no slot for) are all the same omission.
- A test that pins the invariant rather than one substring per clause: `AlterField.from_field_states(t, prev, other).backwards() == AddField.from_field_state(t, prev).forwards().replace("DEFINE FIELD ", "DEFINE FIELD OVERWRITE ", 1)`, parametrised over each `FieldState` attribute. It fails today on every `DEFAULT` case, which is the point. A second one, `for f in fields(FieldState): assert f"previous_{f.name}" in AlterField.__dataclass_fields__`, would have turned #195 into a red test.

### Release note

Migration files already generated on 0.32.6 to 0.33.1 carry no `previous_value` / `previous_encrypted`, so their rollback still emits an OVERWRITE without `VALUE` and nothing warns. Worth a line in the changelog.

### Minor

- The `FieldState` objects in the integration tests leave `nullable=True`, so both directions emit `TYPE option<string>` against a column defined as `TYPE string`; the test cannot see a regression in the type half of the rollback.
- The two integration tests share their setup; the "forward direction really drops hashing" guard is the first test with the `backwards()` call removed and would be free as one sequential test.
- `tests/test_issue_170_unit.py::TestAlterFieldRollbackCompleteness` is already the home for "what does `backwards()` restore"; the new file splits that concern across two modules.
- The 17-line module docstring is changelog narrative that the PR body already carries.

### Checked and fine

ruff, mypy strict, all new tests pass. `from_field_states` is the only producer of `AlterField` in `src/`, `FieldState.__eq__` compares both `encrypted` and `value`, and the generator serialises `previous_encrypted` correctly. #200 is a faithful port and passes on 2.6.5; the `DEFAULT` divergence applies there too.

EulogySnowfall:

Pushed in `d903b49`. All three `DEFAULT` divergences reproduced on the branch first, exactly as written:

```
time::now()  ->  DEFAULT 'time::now()'   (forwards: DEFAULT time::now())
True         ->  DEFAULT True            (forwards: DEFAULT true)
"it's"       ->  DEFAULT 'it's'          parse error
```

### The structural fix, not the third copy

Rather than fix `backwards()`'s DEFAULT branch, the three renderers — `AddField.forwards`, `AlterField.forwards`, `AlterField.backwards` — now delegate to one `_render_define_field` driven off a `FIELD_DEFINITION_FIELDS` tuple, with `backwards()` passing `getattr(self, previous_<k>)`. That is your suggestion taken as written, and it is what makes #170, #195 and this DEFAULT bug one fixed omission instead of three.

One wart worth naming: the previous slot for the type is `previous_type`, not `previous_field_type`. Generated migration files on disk spell it that way, so renaming would break them; `_previous_attr()` carries the exception explicitly rather than silently.

### `previous_encrypted` — removed

You were right that it encodes a state no diff can emit. `from_field_states` resolves `encrypted` into `previous_value` as the argon2 expression, `backwards()` keeps one `if value:` branch, and the two tests that pinned the precedence rule are gone — one replaced by the same assertion built through `from_field_states`, which is the only producer.

### The invariant test — added, and it did its job

```python
AlterField.from_field_states(t, prev, other).backwards()
    == AddField.from_field_state(t, prev).forwards().replace("DEFINE FIELD ", "DEFINE FIELD OVERWRITE ", 1)
```

parametrised over each `FieldState` attribute. It fails on every `DEFAULT` case without the change, as you predicted. The second one — every `FieldState` attribute has a `previous_*` slot — is there too, with `encrypted` named as the one deliberate exception and why.

Adding it surfaced `COMMENT`: `AddField` rendered it, `AlterField` had no slot, and `FieldState` had no attribute at all, so an alter dropped it in both directions. `FieldState.comment` and the parser's COMMENT half close that.

### Minor

- The integration tests left `nullable=True`, so both directions emitted `TYPE option<string>` against a `TYPE string` column. Fixed, and the test now asserts on the stored definition so the type half of the rollback is actually covered.
- The two are one sequential test: the forward direction asserts hashing really was dropped, which is what stops the rollback assertion being vacuous.
- The 17-line docstring is down to the invariant and the consequence.

### One thing I did not do

**I left the new unit file separate from `test_issue_170_unit.py::TestAlterFieldRollbackCompleteness`.** Your concern about splitting "what does `backwards()` restore" across two modules is fair, but that module is already 594 lines and the new file's centre of gravity is different: it pins the *three renderers agreeing*, which is broader than rollback completeness. Merging would produce a 775-line module mixing two concerns rather than consolidating one. Happy to move it if you would still rather they were together.

### Release note

The 0.32.6–0.33.1 caveat is on `AlterField`'s docstring, where someone debugging a rollback will hit it. I deliberately did not add an `## [Unreleased]` section to `CHANGELOG.md`: four open PRs each adding one would conflict, and this repo's convention is that the release PR writes the changelog. Flagging it here so it is picked up there — along with #197's behaviour change.

### Verification

Unit and integration suites green against 3.2.4, mypy strict clean, ruff clean. A new integration test pins the unquoted `DEFAULT time::now()` server-side.

The v2 port is updated in #200; `REFERENCE`/`ON DELETE` is absent there, as it was never backported.

PR #197, opened 2026-09-06
fix(cli): stop makemigrations writing REMOVE TABLE for unmodelled tables

EulogySnowfall:
Third of three PRs from the review of what shipped unreviewed. Independent of #194 and #196.

### The defect

Before 0.33.1 `makemigrations` diffed against an empty `SchemaState`, so `diff()`'s "tables in self but not in target" branch never fired from it. Diffing against the live database made it reachable for the first time — and it fires for **every** table with no registered model. Reproduced against 3.2.4:

```
tables vues en base : ['_surreal_orm_migrations', 'f1_edge', 'f1_only']
DropTable proposés  : ['_surreal_orm_migrations', 'f1_edge']

forwards() : REMOVE TABLE _surreal_orm_migrations;
backwards(): ''                      reversible = False
```

Three categories get swept up, and none of it is recoverable:

- **The ORM's own migration history.** Applying that migration erases the record of what has been applied.
- **`RELATE` edge tables.** `ModelIntrospector` deliberately skips `Relation` / `ManyToMany` fields, so an edge table can *only ever* look unmodelled.
- **Anything whose model module was not imported** — trivially the case with `-m/--models`.

### The fix, at two different levels

**The migrations table is excluded in `diff()`.** Erasing the record of what has been applied is never a reasonable outcome, whatever the caller asked for — so it is an invariant, and no flag turns it back on.

**Everything else stays a drop candidate in `diff()`.** It is a general comparison API and `schema_diff()` should keep reporting the difference. The *policy* belongs to the CLI: `makemigrations` no longer writes those operations into a migration file unless asked, via `--drop-missing`.

Silently dropping them would trade one problem for another, so it lists what it skipped:

```
Skipped 1 table(s) present in the database with no model:
  - legacy_audit
Pass --drop-missing to remove them (irreversible).
```

### Verification

The original reproduction, replayed:

```
tables vues en base : ['_surreal_orm_migrations', 'f1_edge', 'f1_only']
DropTable proposés  : ['f1_edge']
-> _surreal_orm_migrations serait supprimé : False
```

| Command | Result |
| --- | --- |
| CLI suite (with the `cli` extra) | 34 passed |

New: `tests/test_makemigrations_drop_unit.py` (the invariant, plus a guard that other tables *are* still candidates so `diff()` keeps working) and four CLI cases — default skips, skipped tables are reported and the flag named, `--drop-missing` opts in, and the migrations table is not dropped even with the flag.

### Note on `DropField`

Left alone deliberately. A field the model no longer declares is a real schema change the user made, so dropping it is correct. The one exception — the one-time `ManyToMany` / `Relation` columns that existing databases physically carry — is already called out in the 0.33.0 release notes as something to review rather than apply blind.

### Release notes required

- **Fixed** — `makemigrations` wrote an irreversible `REMOVE TABLE` for every database table with no registered model, including the ORM's own migration history and `RELATE` edge tables. The history table is now never a drop candidate, and the rest require `--drop-missing`; skipped tables are listed.
- **Changed** — `makemigrations --drop-missing` is the new opt-in for removing unmodelled tables.

### Still to come

A, B and C on `v2` (0.21.5 carries all three), plus **D** — the gaps in the #191 backport: the JSON-encoder regression, the missing `before_validator`, `merge()` coercing in place, and `bulk_update()` not coercing at all.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,15 @@ warn_unused_configs = true
 module = ["aiohttp.*", "httpx.*", "click.*"]
 ignore_missing_imports = true
 
+# The CLI imports click conditionally: it is an optional extra for users and a
+# dev dependency here. Its decorators therefore type-check differently depending
+# on whether click is installed, so the `type: ignore` comments the module needs
+# when it is absent are reported as unused when it is present. Neither state is
+# wrong, and pinning either one breaks the other environment.
+[[tool.mypy.overrides]]
+module = ["surreal_orm.cli.*"]
+warn_unused_ignores = false
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 python_files = ["test_*.py"]
@@ -109,6 +118,10 @@ include = ["src"]
 
 [dependency-groups]
 dev = [
+    # CI runs `uv sync --frozen`, which does not install the `cli` extra, so
+    # every CLI test skipped there and the commands were only ever exercised
+    # locally. It is a dev dependency here and an optional extra for users.
+    "click>=8.1.8",
     "ipykernel>=7.2.0",
     "mypy>=1.14.0",
     "pytest>=8.3.4",
--- a/src/surreal_orm/cli/commands.py
+++ b/src/surreal_orm/cli/commands.py
@@ -11,8 +11,10 @@ from typing import Any
 
 try:
     import click
-except ImportError:
-    click = None
+except ImportError:  # pragma: no cover - depends on whether the `cli` extra is installed
+    # Annotated as Any so the fallback assignment type-checks in both
+    # environments: with click installed mypy infers Module and rejects None.
+    click = None  # type: ignore[assignment]
 
 
 def require_click() -> None:
@@ -125,7 +127,7 @@ if click is not None:
     @click.option(  # type: ignore[untyped-decorator]
         "--drop-missing",
         is_flag=True,
-        help="Also emit REMOVE TABLE for database tables that have no model (irreversible)",
+        help="Also emit the irreversible removals (REMOVE TABLE/FIELD/ANALYZER/...) for database objects that have no model",
     )
     @click.pass_context  # type: ignore[untyped-decorator]
     def makemigrations(
@@ -139,7 +141,7 @@ if click is not None:
         """Generate migration files from model changes."""
         from ..migrations.generator import MigrationGenerator, generate_empty_migration
         from ..migrations.introspector import introspect_models
-        from ..migrations.operations import DropTable
+        from ..migrations.operations import split_destructive
         from ..migrations.state import SchemaState
 
         migrations_dir = ctx.obj["migrations_dir"]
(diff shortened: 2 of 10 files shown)


EulogySnowfall:
## Review

Right direction, and the "say what was skipped" part is the correct call. The scope is too narrow though: the guard catches one irreversible operation while `diff()` emits several others of exactly the same "the database has it, no model can ever declare it" class.

### Should fix before merge

**Analyzers, APIs and GraphQL config still get an irreversible `REMOVE` by default.** `ModelIntrospector` only ever fills `state.tables`; it never produces `analyzers`, `apis` or `graphql_config`, while `DatabaseIntrospector` does. Reproduced on the branch:

```python
db.analyzers["english"] = AnalyzerState(name="english", tokenizers=["blank"])
db.diff(introspect_models([Users]))
# [DropTable('has_player', reversible=False), RemoveAnalyzer('english', reversible=False)]
```

`makemigrations` skips the first and writes the second into the file on every run, with no "Skipped" line. `migrate` then removes the analyzer that full-text indexes reference, and `rollback` cannot restore it. Same for `DropField` on `in`/`out` of a `TYPE RELATION` table, which SurrealDB 3.2.4 defines itself. `Operation.reversible` already exists on the base class: partitioning on `not op.reversible` behind the one opt-in flag covers all of these instead of enumerating `DropTable` today and the next noun after the next bug report.

### Should fold in

- `schemadiff` and the public `schema_diff()` apply no filter, so the two commands now disagree about the same database: `schemadiff` reports `Drop table has_player` while `makemigrations` says `No changes detected`. A CI gate on `schemadiff` fails permanently on every RELATE edge table. The split belongs one layer down (e.g. `split_destructive(operations)` in the migrations package) so the CLI and library callers share it.
- `from .executor import MIGRATIONS_TABLE` inside `diff()` makes the pure data module depend on the executor (and through it the connection manager) for one string. Define the constant in `state.py` or a small `constants.py` and have `executor.py` import it from there.
- The module docstring of `tests/test_makemigrations_drop_unit.py` contains French (`DropTable proposés`); the repo convention is English.
- The four CLI tests are skipped in CI: `click` lives only in the `cli` extra and CI runs `uv sync --frozen`. The partition logic deserves a plain function tested without click, or `click` belongs in the `dev` group.
- `test_the_migrations_table_is_never_dropped_even_with_the_flag` asserts only that a substring is absent from stdout, with no `exit_code` check, so it passes if the command crashes before describing anything.

### Minor

- `_db_state` in `test_cli.py` and `_db_with` in the new unit file are the same helper; both bypass `SchemaState.add_table()`.
- The rationale is written three times (commands.py comment, test-class docstring, unit-module docstring with a pasted REPL transcript). Once, at the call site, is enough.
- `--drop-missing` combined with `--no-from-db` is accepted silently but can never produce a drop.
- Behaviour change worth a release-note line: since 0.33.1 `makemigrations` produced `REMOVE TABLE` for a deleted model; now it needs `--drop-missing`.

### Checked and fine

ruff, mypy strict, the unit tests pass. `MIGRATIONS_TABLE` is excluded from the drop branch only; the create/modify branches can still act on it if a model declares that name, but that is an edge case. #201 is a faithful port; its CLI tests are skipped on v2 for the same `click` reason.

EulogySnowfall:

Pushed in `654522f`. Reproduced your case first:

```
db.analyzers["english"] = AnalyzerState(name="english", tokenizers=["blank"])
db.diff(introspect_models([Users]))
-> [DropTable('has_player', reversible=False), RemoveAnalyzer('english', reversible=False)]
```

`RemoveAnalyzer` went straight into the file with no "Skipped" line, exactly as reported.

### The partition, not the enumeration

`split_destructive(operations)` partitions on `Operation.reversible`, as you suggested. I checked the flag is actually load-bearing before keying on it: every irreversible operation `diff()` can emit is a removal — `DropTable`, `DropField`, `DropIndex`, `RemoveAccess`, `RemoveAnalyzer`, `RemoveEvent`, `RemoveApi`, `RemoveGraphQLConfig` — and `AlterField`'s conditional `reversible` is always `True` from the diff, since `from_field_states` always carries the previous state. So the rule reads exactly as "what cannot be rolled back is opt-in", including the `in`/`out` `DropField` case you mentioned.

It lives in the migrations package, so `makemigrations` and `schemadiff` share it. `schemadiff` reports the two groups separately now — the disagreement you found (one saying "Drop table has_player", the other "No changes detected") is gone, and a CI gate on it no longer fails permanently on every RELATE edge table.

### Folded in

- `MIGRATIONS_TABLE` moved to `constants.py`; `state.py` no longer imports the executor for one string.
- `--drop-missing` with `--no-from-db` now says it can have no effect.
- The French docstring is gone; the rationale is stated once, at `split_destructive`.
- `_db_state` and `_db_with` are one helper.
- `test_the_migrations_table_is_never_dropped_even_with_the_flag` checks `exit_code` and asserts on the generated migration file. You were right that stdout was the wrong anchor — I applied that to the other drop tests too, since what matters is whether the operation is *written*, not whether a word appears.

### The `click` half — worth reporting

I did both branches of your either/or: the partition is a plain function with direct unit tests, **and** `click` is in the `dev` group.

The second was immediately worth it. With `click` installed the CLI tests actually ran, and one failed on my own change to the "Skipped" message — a break CI would never have caught.

It also exposed a symmetric problem: the conditional import needs its `type: ignore` comments when `click` is absent and mypy reports 38 of them unused when it is present. Pinning either state breaks the other environment, so `warn_unused_ignores` is scoped off for `surreal_orm.cli.*` with the reason recorded. Verified mypy clean **both** with and without `click` installed. If you would rather not carry that override, the fallback is dropping `click` from `dev` and relying on the plain-function tests alone — say the word.

### Release note

The behaviour change — since 0.33.1 `makemigrations` produced `REMOVE TABLE` for a deleted model, now it needs `--drop-missing`, as does every other irreversible removal — is in the commit message and flagged here for the release PR. I did not add an `## [Unreleased]` section: four open PRs each adding one would conflict.

### Verification

Unit and integration suites green against 3.2.4, mypy strict clean in both environments, ruff clean.

The v2 port is updated in #201.

PR #198, opened 2026-09-06
fix(orm): close the gaps left by the ForeignKey coercion backport

EulogySnowfall:
Found by reviewing **#191 after it was merged**. The helpers it carried over are byte-identical to `main`; what was missing is everything `main`'s #174 changed *around* them. Each point below was reproduced on this branch before being fixed.

### A regression the backport introduced

`SurrealJSONEncoder` never learned about `RecordId`. Coercing foreign keys made every `protocol="json"` save or filter fail:

```
TypeError: Object of type RecordId is not JSON serializable
```

This **worked before #191** — a plain `"table:id"` string encodes fine. Coercion is what made the JSON protocol reach a type it did not know.

### The "four interchangeable forms" were one

`_ForeignKeyMarker` and `_ManyToManyMarker` kept a bare `str` schema, without `main`'s `no_info_before_validator_function(record_link_to_str, ...)`. So:

```python
GapHolder(author=GapTarget(id="alice"))
# ValidationError: Input should be a valid string
```

The PR body claimed the model instance, `"table:id"`, a bare ID and a `RecordId` were interchangeable. Only two of them were.

### And that made the next one fatal

`merge()` coerced its data **in place**, so the post-write `setattr` assigned a `RecordId` into a `str` field. With `validate_assignment=True` that raises — *after* the DB write has committed, leaving the caller with an exception and a persisted change. `update()` had the same shape for its signals, which received `RecordId` objects instead of the caller's values.

Both now keep the caller's values for signals and `setattr` and coerce only the copy that goes on the wire, which is how `main` structures it.

### `bulk_update()` was never wired at all

It bound its SET values verbatim, so the original #169 defect survived there untouched: a `"table:id"` string still went into a `record<>` column as a string.

### Verification

The original #174 reproduction still holds (`save()` OK, filter finds the rows), and the three symptoms above are gone.

| Command | Result |
| --- | --- |

### One thing worth recording for future ports

`from src.surreal_sdk...` and `from surreal_sdk...` produce **different class objects**, so `isinstance` silently fails across them. My first run of these tests showed a `TypeError` that looked like a real gap and was purely the import path — production only ever has one. The tests import the way the package does. I verified the distinction explicitly rather than assuming it:

```
RecordId produit par to_record_id -> surreal_sdk.protocol.cbor
encodeur bare  attend             -> surreal_sdk.protocol.cbor      -> {"a": "t:a"}
encodeur src.  attend             -> src.surreal_sdk.protocol.cbor  -> TypeError
```

### Not fixed here

The `FkHolder` fixture in `tests/test_issue_174_v2_unit.py` claims to cover an aliased foreign key and declares no alias, so both alias branches are untested — the very case the original PR called "the dangerous one". Left for the alias-coverage work rather than bolted on, since it needs a fixture with a real alias and a filter test to be worth anything.

### Release notes required

- **Fixed** — foreign-key coercion broke the JSON protocol: any save or filter on an FK raised `TypeError: Object of type RecordId is not JSON serializable` under `protocol="json"`.
- **Fixed** — passing the related model instance (or a `RecordId`) as a foreign-key value raised `ValidationError`; the documented interchangeable forms now all work.
- **Fixed** — `merge()` raised after its write had committed, and `update()` sent coerced values to signals instead of the caller's.
- **Fixed** — `bulk_update()` still wrote foreign keys as strings into `record<>` columns.

### Programme

`v2` still needs A, B and C — the three causes fixed on `main` in #194, #196 and #197, all of which 0.21.5 carries too.

### DIFF
--- a/src/surreal_orm/fields/relation.py
+++ b/src/surreal_orm/fields/relation.py
@@ -28,6 +28,7 @@ from pydantic import GetCoreSchemaHandler, GetJsonSchemaHandler
 from pydantic.json_schema import JsonSchemaValue
 from pydantic_core import CoreSchema, core_schema
 
+from surreal_orm.model_base import record_link_to_str
 from surreal_orm.utils import escape_record_id
 
 if TYPE_CHECKING:
@@ -129,15 +130,22 @@ class _ForeignKeyMarker:
                 related_name = arg.related_name
                 break
 
+        # A before-validator renders the accepted forms — model instance,
+        # RecordId, "table:id" — down to the string the schema expects. Without
+        # it the "four interchangeable forms" are one, and merge()'s post-write
+        # setattr of a coerced value raises.
         return core_schema.nullable_schema(
-            core_schema.str_schema(
-                metadata={
-                    "relation_type": "foreign_key",
-                    "to_model": to_model,
-                    "on_delete": on_delete,
-                    "related_name": related_name,
-                    "surreal_type": "record",
-                }
+            core_schema.no_info_before_validator_function(
+                record_link_to_str,
+                core_schema.str_schema(
+                    metadata={
+                        "relation_type": "foreign_key",
+                        "to_model": to_model,
+                        "on_delete": on_delete,
+                        "related_name": related_name,
+                        "surreal_type": "record",
+                    }
+                ),
             )
         )
 
@@ -193,8 +201,12 @@ class _ManyToManyMarker:
                 break
 
         # ManyToMany is represented as a list of record IDs (virtual field)
+        # Each item accepts a model instance, RecordId, or "table:id" string.
         return core_schema.list_schema(
-            core_schema.str_schema(),
+            core_schema.no_info_before_validator_function(
+                record_link_to_str,
+                core_schema.str_schema(),
+            ),
             metadata={
                 "relation_type": "many_to_many",
                 "to_model": to_model,
--- a/src/surreal_orm/model_base.py
+++ b/src/surreal_orm/model_base.py
@@ -193,13 +193,13 @@ def record_link_to_str(value: Any) -> Any:
         record_link_to_str(RecordId(table="users", id="a"))  # "users:a"
         record_link_to_str("users:alice")                    # "users:alice"
     """
-    # Duck-typed like _convert_record_id_to_string: avoids import path
-    # issues between src.surreal_orm and surreal_orm. The isinstance guard
-    # keeps unrelated objects with a plain `record_id` attribute untouched.
+    # `record_id` is checked first: a model instance carries one, and its own
+    # class is not a RecordId. The predicate keeps an unrelated object with a
+    # plain `record_id` attribute from being mistaken for a link.
     record_id = getattr(value, "record_id", None)
-    if isinstance(record_id, RecordId):
+    if is_record_link(record_id):
         return str(record_id)
-    if isinstance(value, RecordId):
+    if is_record_link(value):
         return str(value)
     if record_id is None and callable(getattr(value, "get_id", None)):
         raise ValueError(f"cannot reference an unsaved {type(value).__name__}; save it first")
@@ -221,14 +221,15 @@ def _to_record_link(value: Any, table: str | None) -> Any:
     Returns:
         A ``RecordId``, or the original value
     """
-    if isinstance(value, RecordId):
+    if is_record_link(value):
         return value
 
     value = record_link_to_str(value)
 
     # A bare ID is only resolvable against a known target table. "$var"
     # references stay untouched — they are user-supplied query variables.
-    if isinstance(value, str) and table and not value.startswith("$"):
+    # An empty string is not an id: qualifying it produces RecordId(table, "").
+    if isinstance(value, str) and value and table and not value.startswith("$"):
         value_table, id_part = parse_record_id(value)
         if value_table is None:
             return RecordId(table=table, id=id_part)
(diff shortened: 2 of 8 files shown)


EulogySnowfall:
## Review

Two of the four fixes are right (the before-validator on the FK markers, and wiring `bulk_update`). The other two need another pass, and the bulk_update change has no test coverage. Reproduced on the branch; the server-side point against `surrealdb/surrealdb:v2.6.5` with `protocol="json"`.

### Blocking

**1. The JSON encoder loses the ID's type.** `str(RecordId)` emits `table:id` with no escaping. A string ID `"123"` becomes the numeric record over JSON, while CBOR writes the string one:

```
json:  CREATE h SET a = $a  with RecordId("t", "123")  ->  a = t:123 AS is_numeric_id: true,  a = t:⟨123⟩: false
cbor:  same call                                        ->  a = t:⟨123⟩: true
```

Two different records, no error, and the JSON path was the one this PR set out to repair. `format_thing` already knows how to escape; the encoder should emit the same form.

**2. `TestBulkUpdateCoercesForeignKeys` never calls `bulk_update()`.** It exercises `_to_record_link` directly and ends with `assert qs is not None`. Removing the new block in `bulk_update` leaves the suite green. `tests/test_permission_denial_unit.py::_patch_client` has the mocking idiom: patch `get_client`, call `bulk_update({"author": "alice"})`, assert `variables["_bu0"]` is a `RecordId`.

### Should fix

- **`merge()` tx path leaves the local instance unqualified.** The setattr loop now iterates the caller's raw `data_set`: `merge(author="alice", tx=tx)` writes `gap_targets:alice` but leaves `holder.author == "alice"`, while the non-tx path refreshes to `"gap_targets:alice"`. `main` iterates `wire_data`. The comment justifying the change ("assigned a RecordId back onto a str field, which raised") is no longer true once the validator accepts a `RecordId`, so `wire_data` can be used here as on `main`.
- `update()` and `merge()` log the uncoerced dict while the wire carries `wire_data`; `main` logs `wire_data`. `QueryLogger` then shows the exact string payload the DB rejects and hides that coercion happened.
- The second JSON encoder, `_SurrealJSONEncoder` in `utils.py`, was not updated, so `raw_query(..., inline_dicts=True)` with a `record_id` inside the dict raises `Failed to serialize variable`. One encoder (the SDK's) would do; note an inlined record link should be an unquoted thing, not a JSON string.
- `_to_record_link("", "users")` returns `RecordId(table='users', id='')`. An empty string should not be qualified.

### Minor

- `data_set = {key: value for key, value in data.items()}` is a copy of a dict that is already private to the call, and `_coerce_foreign_keys` copies again.
- `record_link_to_str` uses `isinstance` against the `surreal_sdk` `RecordId` while the neighbouring `_convert_record_id_to_string` is deliberately duck-typed for the `src.` import split the PR body itself describes. One predicate for "this is a record link" would be better than three.
- The new comments narrate the bug ("was simply never wired here", "raises TypeError") rather than the invariant; the PR body and test docstring already carry the history. The PR body also contains a French terminal transcript.
- `GapTarget`/`GapHolder` duplicate `FkTarget`/`FkHolder` from `test_issue_174_v2_unit.py`.

### Checked and fine

ruff, mypy strict, the unit suite passes. No circular import: `surreal_orm/__init__` loads `fields` before `model_base`, and `model_base`'s imports of `fields.relation` are lazy. The original #174 reproduction still holds.

EulogySnowfall:

Pushed in `4c31b04`. Both blocking points reproduced against `surrealdb:v2.6.5` first.

### 1. The JSON encoder lost the id's type

Confirmed, and I checked what the server actually does with the two forms:

```
CREATE t:123 …   -> id t:123    meta::id(id) IS 123 -> true
CREATE t:`123` … -> id t:`123`  meta::id(id) IS 123 -> false
```

Two distinct rows. `RecordId.to_surql()` escapes the way `format_thing` does, and the encoder uses it.

Where it lives matters: `format_thing` is in `surreal_orm.utils`, which the SDK cannot import. Rather than a second copy, the escaping moved **into** the SDK — where the wire format needs it — and `surreal_orm.utils` re-exports it. Net one implementation instead of two.

### 2. `TestBulkUpdateCoercesForeignKeys` never called `bulk_update()`

Confirmed by deleting the coercion: the suite stayed green. It now uses the `_patch_client` idiom you pointed at, calls `bulk_update({"author": "alice"})` and asserts `variables["_bu0"]` is a `RecordId`. Verified red without the fix.

### Should fix — all four

- **`merge()` tx path.** Fixed; it iterates `wire_data`. You were right that the comment justifying the old behaviour stopped being true once the validator accepted a `RecordId` — I removed the comment along with the behaviour.
- **Logging.** `update()` and `merge()` log `wire_data`, so `QueryLogger` shows what the database actually received.
- **`_SurrealJSONEncoder`.** Fixed, and your parenthetical was the important half: an inlined record link has to be an unquoted thing, not a JSON string, or it is stored as text. It goes through a sentinel that is stripped after `dumps()`.

  That surfaced a second defect on the same line: `json_str` was passed to `re.sub` as a replacement *template*, so a backslash in the JSON was read as a group reference — the v2 side of #193. Fixed here since this PR is what put a `\`-bearing value on that path.
- **`_to_record_link("", …)`.** Fixed; an empty string is returned unchanged.

### Minor

- The redundant `data_set` copy is gone.
- One `is_record_link()` predicate replaces the three spellings. Worth noting why `isinstance` alone cannot be it: `surreal_sdk` and `src.surreal_sdk` are two distinct classes at runtime, which is what the duck-typing was working around — so the predicate keeps the module check and adds the attribute check, rather than picking one.
- The narrating comments are down to the invariant.
- `GapTarget`/`GapHolder` are gone; the tests import `FkTarget`/`FkHolder`.

### Added

An integration test proving the JSON path server-side: a string id of `"123"` must not land as the numeric record. It fails with `str()` in place of `to_surql()`, so it pins the actual defect rather than the encoder's output shape.

### Verification

Unit suite and 372 integration tests green against 2.6.5, mypy strict clean, ruff clean.

### Not addressed

The French terminal transcript in the PR description — I have left the body alone rather than rewrite it under you. Happy to if you would like.

PR #199, opened 2026-09-06
fix(migrations): make a table definition survive the round trip

EulogySnowfall:
Backport of **#194** onto the v2 LTS line, re-implemented against this branch. All three defects are present here verbatim, and each was reproduced on a live SurrealDB **2.6.5** before porting.

### 1. Declared table permissions never reached the database

```
DEFINE TABLE p SCHEMAFULL;                                 -> OK
DEFINE TABLE p PERMISSIONS FOR select WHERE $auth.id = id; -> ERR "The table 'p' already exists"
in the database: DEFINE TABLE p TYPE NORMAL SCHEMAFULL PERMISSIONS NONE
```

A model declaring `permissions={"select": "$auth.id = id"}` got a table with **no** rule. It fails closed (`PERMISSIONS NONE` denies record users), so this is broken access control rather than leaked data — but the rule the user asked for was never applied. And since #187 made the executor surface per-statement errors, the *first* migration of any such model now aborts.

`PERMISSIONS` is a clause of `DEFINE TABLE`; it is emitted as one.

### 2. Rolling back a table change dropped the table

The diff reuses `CreateTable` to redefine an existing table, and its `backwards()` was `REMOVE TABLE`. A permissions-only change rolled back by dropping the table and every row in it, while reporting `reversible = True`.

`CreateTable` gains an **overwrite mode** — `DEFINE TABLE OVERWRITE`, which the plain form's rejection made necessary anyway — and carries the previous definition so `backwards()` restores it. `_render()` is shared by both directions, so a rollback cannot drift from the definition it restores. A table the migration actually *created* still rolls back to `REMOVE TABLE`.

### 3. The permissions the server reports could not be parsed back

```
PERMISSIONS FOR select WHERE $auth.id = id, FOR create, update, delete NONE
```

The condition kept the comma separating the groups, and the keyword form has no `WHERE`, so those three actions were dropped entirely. **This one is load-bearing**: without it the first fix would only make the mismatch *visible* — permissions would finally land, and every `makemigrations` would then propose a redundant redefinition forever.

### Verification

| Command | Result |
| --- | --- |

11 unit cases and 4 integration, including one that writes a row, applies the redefinition, rolls it back, and asserts the row is still there.

**One adaptation from the main version:** the integration tests build the DDL from `CreateTable` / `AddField` directly rather than through `define_table()`, which does not exist on this branch. That is what the migration executor does anyway, so it exercises the same path.

### Release notes required

- **Fixed** — table permissions declared in `model_config` never reached the database: the clause was emitted as a second `DEFINE TABLE`, which SurrealDB rejects for an existing table.
- **Fixed** — a migration that changed a table definition rolled back with `REMOVE TABLE`, dropping the table and all its rows while reporting itself reversible.
- **Fixed** — the `PERMISSIONS` shape `INFO FOR DB` reports could not be parsed back, so a permissioned table diffed as changed on every run.

### Programme

`v2` backports from the review: **A (this) → B → C**. #198 (the FK-coercion gaps) is independent and already open.

### DIFF
--- a/src/surreal_orm/migrations/__init__.py
+++ b/src/surreal_orm/migrations/__init__.py
@@ -31,6 +31,7 @@ from .model_generator import ModelCodeGenerator
 from .operations import (
     AddField,
     AlterField,
+    AlterTable,
     CreateIndex,
     CreateTable,
     DataMigration,
@@ -57,6 +58,7 @@ __all__ = [
     # Migrations
     "Migration",
     "Operation",
+    "AlterTable",
     "CreateTable",
     "DropTable",
     "AddField",
--- a/src/surreal_orm/migrations/db_introspector.py
+++ b/src/surreal_orm/migrations/db_introspector.py
@@ -166,6 +166,7 @@ class DatabaseIntrospector:
             view_query=table_props.get("view_query"),
             relation_in=table_props.get("relation_in"),
             relation_out=table_props.get("relation_out"),
+            comment=table_props.get("comment"),
             enforced=table_props.get("enforced", False),
         )
 
--- a/src/surreal_orm/migrations/define_parser.py
+++ b/src/surreal_orm/migrations/define_parser.py
@@ -26,6 +26,10 @@ _FIELD_KEYWORDS = [
     "TYPE",
 ]
 
+# The four actions a table PERMISSIONS clause can govern, in the order
+# SurrealDB reports them.
+_PERMISSION_ACTIONS = ("select", "create", "update", "delete")
+
 # Keywords that delimit clauses in a DEFINE TABLE statement.
 _TABLE_KEYWORDS = [
     "SCHEMAFULL",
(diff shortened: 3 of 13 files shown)


EulogySnowfall:
## Review

Faithful backport of #194: the only differences from the `main` diff are hunk offsets and the branch adaptations described in the body. The test files this PR adds pass against `surrealdb/surrealdb:v2.6.5`.

Every finding in the review of #194 applies here unchanged. I re-ran the server-side reproductions on 2.6.5 where relevant and they behave the same as on 3.2.4. Please fix on `main` first and re-port, so the two branches stay in lockstep.

Full review: https://github.com/EulogySnowfall/SurrealDB-ORM/pull/194

EulogySnowfall:

As requested: fixed on #194 first, then re-ported here in `67e8ca1`. Every point was **re-verified against `surrealdb:v2.6.5` on this branch**, not carried over on the assumption that main's reproduction transfers.

The full point-by-point response is on #194. What differs here:

- **`define_table()` does not exist on this branch**, so its OVERWRITE and per-statement-status changes have no target. The equivalent invariant — the operations the executor emits can be applied twice — is covered through the port's existing `_apply_model_schema()` helper instead.
- Everything else ports unchanged: `FULL`/`NONE` as keywords, the `TYPE USER` parse error (2.6.5 rejects it identically), the bare-overwrite reset, the flattened view, the quote-aware and no-longer-quadratic permissions split, the `meta.type` truncation, `COMMENT` in both directions, and the collapse onto `from_table_state()` / `from_table_states()`.

PR #200, opened 2026-09-06
fix(migrations): restore the VALUE clause when rolling back an AlterField

EulogySnowfall:
Backport of **#196** onto the v2 LTS line. The gap is identical here — `previous_value` and `previous_encrypted` appear nowhere on either branch.

### The defect

`AlterField.backwards()` re-emits `DEFINE FIELD OVERWRITE`, which replaces the **whole** definition, so every clause without a `previous_*` slot is silently dropped:

```
forwards()  -> DEFINE FIELD OVERWRITE password ON users TYPE string;
backwards() -> DEFINE FIELD OVERWRITE password ON users TYPE string;
```

Rolling that back drops `VALUE crypto::argon2::generate($value)`. The column **stops hashing**, and every write after the rollback stores the password in plaintext. `Computed` fields lose their expression the same way.

**Timing differs from `main`.** The gap predates the `OVERWRITE` change on both lines, but it was harmless while a rollback was a silent no-op. On `main` it went live with 0.32.6; here it went live with **0.21.5** — the release that carried the `OVERWRITE` backport (#187). Fixing the no-op is what made a latent defect destructive, and nothing in the suite noticed on either line.

### The fix

`previous_value` and `previous_encrypted`, wired by `from_field_states()` — the single place a `FieldState` becomes operation arguments — and emitted by `backwards()` in the same position and with the same precedence as `forwards()` (`encrypted` wins over a plain `value`).

### Verification

Against a live **SurrealDB 2.6.5**: apply `forwards()`, then `backwards()`, then write to the column — the stored value comes back as an `$argon2` hash rather than the plaintext.

A second integration test asserts that `forwards()` **alone** really does store plaintext. Without it the first could pass for the wrong reason.

| Command | Result |
| --- | --- |

### Out of scope, same as on main

Field-level `PERMISSIONS` are not modelled on either branch, so `OVERWRITE` still wipes any set outside the ORM. Tracked in **#195** — it affects `forwards()` as much as `backwards()`, so no `previous_*` slot can fix it.

### Release notes required

- **Fixed** — rolling back a migration that altered an `Encrypted` column dropped its `VALUE crypto::argon2::generate($value)` clause, so the column stopped hashing and later writes stored passwords in plaintext. `Computed` fields lost their expression the same way. Live on this branch since 0.21.5.

### Programme

`v2` backports: **A (#199) → B (this) → C**. Both touch `operations.py`, so whichever merges second will need a rebase.

### DIFF
--- a/src/surreal_orm/migrations/define_parser.py
+++ b/src/surreal_orm/migrations/define_parser.py
@@ -207,6 +207,15 @@ def parse_define_field(statement: str) -> FieldState:
     # Parse ASSERT clause
     assertion = clauses.get("ASSERT") or None
 
+    # Parse COMMENT clause — echoed back single-quoted, with '' for a quote
+    comment: str | None = None
+    raw_comment = clauses.get("COMMENT")
+    if raw_comment:
+        raw_comment = raw_comment.strip()
+        if raw_comment.startswith("'") and raw_comment.endswith("'") and len(raw_comment) >= 2:
+            raw_comment = raw_comment[1:-1]
+        comment = raw_comment.replace("''", "'")
+
     return FieldState(
         name=field_name,
         field_type=field_type,
@@ -217,6 +226,7 @@ def parse_define_field(statement: str) -> FieldState:
         flexible=flexible,
         readonly=readonly,
         value=value_expr,
+        comment=comment,
     )
 
 
--- a/src/surreal_orm/migrations/operations.py
+++ b/src/surreal_orm/migrations/operations.py
@@ -701,6 +701,8 @@ class AlterField(Operation):
     previous_flexible: bool = False
     previous_readonly: bool = False
     previous_nullable: bool = False
+    previous_value: str | None = None
+    previous_encrypted: bool = False
 
     @classmethod
     def from_field_states(cls, table: str, current: "FieldState", target: "FieldState") -> "AlterField":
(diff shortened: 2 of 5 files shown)


EulogySnowfall:
## Review

Faithful backport of #196: the only differences from the `main` diff are hunk offsets and the branch adaptations described in the body. The test files this PR adds pass against `surrealdb/surrealdb:v2.6.5`.

Every finding in the review of #196 applies here unchanged. I re-ran the server-side reproductions on 2.6.5 where relevant and they behave the same as on 3.2.4. Please fix on `main` first and re-port, so the two branches stay in lockstep.

Full review: https://github.com/EulogySnowfall/SurrealDB-ORM/pull/196

EulogySnowfall:

As requested: fixed on #196 first, then re-ported here in `03b3119`. Every point was **re-verified against `surrealdb:v2.6.5` on this branch**, not carried over on the assumption that main's reproduction transfers.

The full point-by-point response is on #196. What differs here:

- **`REFERENCE` / `ON DELETE` is a SurrealDB 3.0 clause and was never backported** (`ReferencesField` does not exist here), so it is absent from `FIELD_DEFINITION_FIELDS`, from the shared renderer and from both dataclasses. The parser's `COMMENT` half is carried; its `REFERENCE` half is not.
- Everything else ports unchanged: the three `DEFAULT` divergences, the single `_render_define_field`, the removal of the derived `previous_encrypted`, the `COMMENT` slot, and the invariant test parametrised over each `FieldState` attribute.

PR #201, opened 2026-09-06
fix(cli): stop makemigrations writing REMOVE TABLE for unmodelled tables

EulogySnowfall:
Backport of **#197** onto the v2 LTS line. The code is identical here, so the defect is too.

### The defect

Before 0.21.5, `makemigrations` diffed against an empty `SchemaState`, so `diff()`'s "tables in self but not in target" branch never fired from it. Diffing against the live database made it reachable for the first time — and it fires for **every** table with no registered model:

```
DropTable proposés : ['_surreal_orm_migrations', 'legacy_audit']
forwards() : REMOVE TABLE _surreal_orm_migrations;
backwards(): ''                      reversible = False
```

Three categories, none recoverable:

- **The ORM's own migration history** — applying that migration erases the record of what has been applied.
- **`RELATE` edge tables** — `ModelIntrospector` deliberately skips `Relation` / `ManyToMany` fields, so an edge table can *only ever* look unmodelled.
- **Anything whose model module was not imported**, trivially the case with `-m/--models`.

### The fix, at two different levels

**The migrations table is excluded in `diff()`.** Erasing the record of what has been applied is never a reasonable outcome, whatever the caller asked for — so it is an invariant, and no flag turns it back on.

**Everything else stays a drop candidate in `diff()`.** It is a general comparison API and `schema_diff()` should keep reporting the difference. The *policy* belongs to the CLI: `makemigrations` no longer writes those operations into a migration file unless asked, via `--drop-missing`.

Silently filtering them would trade one problem for another, so it says what it skipped:

```
Skipped 1 table(s) present in the database with no model:
  - legacy_audit
Pass --drop-missing to remove them (irreversible).
```

### Verification

| Command | Result |
| --- | --- |
| CLI suite (with the `cli` extra) | 34 passed |

### Note on `DropField`

Left alone deliberately, as on `main`. A field the model no longer declares is a real schema change the user made, so dropping it is correct.

### Release notes required

- **Fixed** — `makemigrations` wrote an irreversible `REMOVE TABLE` for every database table with no registered model, including the ORM's own migration history and `RELATE` edge tables. The history table is now never a drop candidate, and the rest require `--drop-missing`; skipped tables are listed.
- **Changed** — `makemigrations --drop-missing` is the new opt-in for removing unmodelled tables.

### Programme

This completes the `v2` side: **A (#199), B (#200), C (this)**, plus **D (#198)**. A and C both touch `state.py`; A and B both touch `operations.py` — whichever merges second will need a rebase.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,15 @@ warn_unused_configs = true
 module = ["aiohttp.*", "httpx.*", "click.*"]
 ignore_missing_imports = true
 
+# The CLI imports click conditionally: it is an optional extra for users and a
+# dev dependency here. Its decorators therefore type-check differently depending
+# on whether click is installed, so the `type: ignore` comments the module needs
+# when it is absent are reported as unused when it is present. Neither state is
+# wrong, and pinning either one breaks the other environment.
+[[tool.mypy.overrides]]
+module = ["surreal_orm.cli.*"]
+warn_unused_ignores = false
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 python_files = ["test_*.py"]
@@ -110,6 +119,10 @@ include = ["src"]
 
 [dependency-groups]
 dev = [
+    # CI runs `uv sync --frozen`, which does not install the `cli` extra, so
+    # every CLI test skipped there and the commands were only ever exercised
+    # locally. It is a dev dependency here and an optional extra for users.
+    "click>=8.1.8",
     "ipykernel>=7.2.0",
     "mypy>=1.14.0",
     "pytest>=8.3.4",
--- a/src/surreal_orm/cli/commands.py
+++ b/src/surreal_orm/cli/commands.py
@@ -11,8 +11,10 @@ from typing import Any
 
 try:
     import click
-except ImportError:
-    click = None
+except ImportError:  # pragma: no cover - depends on whether the `cli` extra is installed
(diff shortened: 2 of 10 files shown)


EulogySnowfall:
## Review

Faithful backport of #197: the only differences from the `main` diff are hunk offsets and the branch adaptations described in the body. The test files this PR adds pass against `surrealdb/surrealdb:v2.6.5`.

Every finding in the review of #197 applies here unchanged. I re-ran the server-side reproductions on 2.6.5 where relevant and they behave the same as on 3.2.4. Please fix on `main` first and re-port, so the two branches stay in lockstep.

Full review: https://github.com/EulogySnowfall/SurrealDB-ORM/pull/197

EulogySnowfall:

As requested: fixed on #197 first, then re-ported here in `194bf65`. Every point was **re-verified against `surrealdb:v2.6.5` on this branch**, not carried over on the assumption that main's reproduction transfers.

The full point-by-point response is on #197. What differs here:

- **`RemoveGraphQLConfig` does not exist here**, so it is absent from the exports. That costs nothing: `split_destructive` keys on `Operation.reversible` rather than on a list of classes, so it covers whatever this branch's `diff()` can emit — verified, `RemoveAnalyzer` is held back here as it is on main.
- `click` in the `dev` group and the scoped `warn_unused_ignores` port as-is; mypy verified clean on this branch both with and without `click` installed.

